PR #7, opened 2016-12-15
Update to HEAD

wesm:


wesm:
@ocefpaf something wrong with Circle CI today?

ocefpaf:
`CircleCI` changed their policy on tests from forks. You will need to navigate to

https://circleci.com/gh/wesm/arrow-cpp-feedstock

and click on `follow project`. After that just send an empty commit here and CircleCI should start. Unfortunately that is not something we can automate. We are working with Circle to see if we can get the old behavior back.

PR #12, opened 2017-01-06
Update to HEAD

xhochy:


xhochy:
Conda currently lacks optional requirements and as the `jemalloc` requirement is quite small, it made it a real requirement: https://github.com/conda/conda/issues/793

wesm:
@xhochy do you know why the OS X build is failing?

xhochy:
jemalloc on OSX in the latest version needed an `#define`. Fixed upstream in `arrow`, now the build is green.

wesm:
I believe there are some API changes that may break pyarrow, so we'll want to update the parquet-cpp and pyarrow packages as soon as possible

xhochy:
Will do immediately!

PR #15, opened 2017-01-30
Update to HEAD (2017-01-29), re-render with conda-smithy 2.0.1

wesm:


wesm:
@ocefpaf do we only have 5 concurrent OS X builds on Travis CI right now? I noticed that there's a gdal build running, and based on prior builds (https://travis-ci.org/conda-forge/gdal-feedstock/builds/195528816), we're looking at at least 45 minutes to an hour of waiting until workers free up.

Perhaps we could limit concurrent workers on a per project basis to 2 or something like that?

ocefpaf:
> @ocefpaf do we only have 5 concurrent OS X builds on Travis CI right now?

Yes.

> I noticed that there's a gdal build running, and based on prior builds (https://travis-ci.org/conda-forge/gdal-feedstock/builds/195528816), we're looking at at least 45 minutes to an hour of waiting until workers free up.
Perhaps we could limit concurrent workers on a per project basis to 2 or something like that? 

Not sure if that is possible but I will take a look. (I do try to send gdal updates only on weekends for that reason btw.)

PR #18, opened 2017-04-30
Update to HEAD

wesm:


wesm:
I'm planning to add Windows builds to the pyarrow-feedstock to make sure that works before we release 0.3.

PR #27, opened 2017-06-19
Retrigger due to jemalloc ABI breakage

xhochy:
cc @jreback


xhochy:
PRs inside conda-forge repos doesn't seem like a good idea.

wesm:
Yes I've been admonished for not using my fork before

PR #28, opened 2017-06-19
Retrigger due to jemalloc ABI breakage

xhochy:


isuruf:
You should pin `jemalloc` so that it doesn't break again.

wesm:
Can we do that?

isuruf:
What's the exact ABI breakage? https://abi-laboratory.pro/tracker/timeline/jemalloc/ suggests v5.0.0 is backwards compatible

You can do `jemalloc >= 5.0.0` in the recipe to make sure the package installs with newer jemalloc.

xhochy:
There is no breakage in jemalloc, it rather is a problem with glibc.  Tactual problem here is caused by a bug in glibc (see fix here: https://www.sourceware.org/git/gitweb.cgi?p=glibc.git;a=commitdiff;h=67cbf9a2a9e8953aa7ea875938d6646727c71aaa).

PR #30, opened 2017-06-28
Update build for 0.5.x given toolchain changes

wesm:


wesm:
cc @bmabey; this is step 1, now I'll update parquet-cpp

PR #35, opened 2017-07-27
Switch off jemalloc allocator pending resolution of ARROW-1282

wesm:


wesm:
Builds failed due to unrelated issues...

PR #39, opened 2017-09-28
MNT: Re-rendered with conda-smithy 2.4.2

xhochy:
Rerendering to see if the numpy version updates.


xhochy:
Merging although Smithy is not happy as the previous commit without the NumPy<->Windows specialization made Smithy happy.

PR #44, opened 2018-01-24
Update boost pin

jschueller:


jschueller:

xhochy:
This sadly fails on Windows.

@MaxRis would be nice if you could have a short look. Maybe you see why boost_filesystem is not found.

MaxRis:
@xhochy best guess, it's because "boost-cpp 1.66.0" is requested with this changes and bld.bat relies on VS 2015 (vc14_0)  as build tool in this line: "cmake -G "Visual Studio 14 2015 Win64"". But boost-cpp package for VS 2015 (vc14_0) is not available in conda. Currently, `boost-cpp:      1.66.0-vc14_1 conda-forge [vc14]` is installed, which is built with VS 2017 (vs14_1). That's why, seems, headers are found fine, but libs are, probably, incompatible and errors out on linkage stage.

jschueller:
No, in vc14_1 , 1 is the build number, not vs 2017.

MaxRis:
@jschueller looks like, I've checked dependencies of dlls installed with vc14_1 and they linked with "MSVCP140.DLL"

jschueller:
Yup, we dont build anything against vs2017 for now (maybe for python 3.7)

xhochy:
Fixed via #45

PR #45, opened 2018-01-24
Use environment variables for VS

xhochy:
cc @MaxRis


xhochy:
@conda-forge-admin please rerender

recipe/bld.bat
```diff
@@ -3,7 +3,7 @@ pushd "%SRC_DIR%"\cpp\build
 
 set ARROW_BUILD_TOOLCHAIN=%LIBRARY_PREFIX%
 
-cmake -G "Visual Studio 14 2015 Win64" ^
+cmake -G "Visual Studio %VS_MAJOR% %VS_YEAR% Win64" ^
```
isuruf: `cmake -G "%CMAKE_GENERATOR%" ^`

xhochy: @isuruf ah, thanks. That is even better :)

MaxRis:
Hey @xhochy ,
That might be related to the following message from the log:
"liblz4_static.lib(lz4.obj) : MSIL .netmodule or module compiled with /GL found; restarting link with /LTCG; add /LTCG to the..."
That's directly related to what I've worked on and not yet finished. I will try to check if it fails exact because of this. Possibly, update of lz4 and zstd feedstocks is required here, since there libs shoudn't be compiled with /GL flag.

MaxRis:
Update: I was able to repro it locally. Most of all it's because of lz4 lib as described above. I'm going to try to update lz4-c-feedstock to use latest lz4-c release (it already contains my changes to not build with /GL flag, so we will not need to patch lz4-c-feedstock anymore).

MaxRis:
@xhochy @jschueller 
It seems that FindBoost of cmake is not yet updated to support changed naming schema of boost's libs starting from version 1.66 https://stackoverflow.com/a/48191548
that should be the reason of the `LINK : fatal error LNK1104: cannot open file 'boost_filesystem.lib'` error.
boost_filesystem.lib is a part of shared libs linkage process on Windows and it shouldn't appear at all, because we're trying to use only static boosts libs ( `libboost_filesystem.lib` ).

jschueller:
Our cmake is already patched, maybe the changes are not enough for windows.

isuruf:
Can you try doing `set "CPPFLAGS=/DBOOST_ALL_NO_LIB"` at the top?

MaxRis:
@jschueller thanks, I will check FindBoost to figure out what exactly doesn't work.

MaxRis:
@isuruf I've tried to set it in terminal globally before starting cmake build, but, seems, no effect.

Also have tried to get latest version of FindBoost.cmake https://raw.githubusercontent.com/Kitware/CMake/master/Modules/FindBoost.cmake and run with it, but still the same issue.

MaxRis:
update: I'm trying to create minimal FindBoost sample with linkage failure error on Windows. And it seems it fails for me so far with similar error as well.

MaxRis:
Was able to repro the same issue on latest cmake nightly build binaries. Have created issue and attached minimal project with issue https://gitlab.kitware.com/cmake/cmake/issues/17717

MaxRis:
@xhochy arrow-cpp-feedstock, finally, is built fine https://ci.appveyor.com/project/MaxRisuhin/arrow-cpp-feedstock
Have created PR for Arrow with tested changes - https://github.com/apache/arrow/pull/1567
Also was needed to pin snappy to 1.1.7 and brotli to 1.0.2 to avoid related compilation issues of arrow-cpp-feedstock (not sure why latest version of libs is not obtained automatically) https://github.com/MaxRis/arrow-cpp-feedstock/commit/1a7ddc5a478fa928c368249952c3ff69faedec11

xhochy:
@MaxRis Can you make a patch so that we can update arrow-cpp=0.8.0 or should we wait for the 0.9.0 release?

MaxRis:
@xhochy I will go for the patch

xhochy:
@conda-forge-admin please rerender

PR #49, opened 2018-03-26
PERF: Fix perf regression

cpcloud:


cpcloud:
@wesm does this look ok to you?

PR #50, opened 2018-03-27
CLN: Remove patch

cpcloud:
It's less fragile to use a commit hash here than to use the patch mechanism.


xhochy:
@cpcloud The checksum did not match. I have updated to what Conda told me, hopefully we don't have anything in our sources again that change based on commits.

cpcloud:
thanks @xhochy

tkelman:
Why is this "less fragile"? You aren't building 0.9.0 with a fix for a single issue any more here, you're building some intermediate unreleased commit partway between 0.9.0 and whatever the next release is.

cpcloud:
Maybe "less fragile" is not the right set of words. How about "less clunky because we don't have to carry a patch around"? I'm happy to iterate on the wording until you're satisfied.

tkelman:
It's the same patch, but 0.9.0 == c695a5ddc8d26c977b5ecd0c55212e900726953e, and 0.9.0 + one patch is not the same thing as f9f8320339692d4134d1ef42c32cb7c8d547593e. The latter is no longer 0.9.0, it's 0.9.0 + https://github.com/apache/arrow/compare/c695a5ddc8d26c977b5ecd0c55212e900726953e...f9f8320339692d4134d1ef42c32cb7c8d547593e, which is 23 commits, no longer a release but a snapshot somewhere in between 0.9.0 and 0.10.0, and I'd consider it a little unexpected to be packaging and releasing that via the main conda-forge channel as a "build number increment" of the package version still called 0.9.0.

PR #54, opened 2018-07-02
MNT: Re-rendered with conda-smithy 3.1.6 and pinning 2018.07.01

xhochy:


xhochy:
It's a bit suspicious that there is only a single CircleCI job. I'll wait 10min and then maybe amend the commit to retrigger the CI notifications.

xhochy:
Better now :)

cpcloud:
Ok the OS X jobs have *still* not started. I will periodically check in, but at this rate it'll probably morning in Europe or the USA before this gets merged.

PR #69, opened 2019-01-10
Add run_exports

### LINKED ISSUE #58: Add run_exports
We may want to add run_exports here (as in AnacondaRecipes) so that ensuring abi compatibility is a bit simpler for downstream applications.

```
build:
  run_exports:
    - {{ pin_subpackage("arrow-cpp", max_pin="x.x") }}
```

Alternatively we may want to merge some of the changes from AnacondaRecipes

xhochy:
Fixes #58 

@mariusvniekerk Is this enough for the pinning? I don't see any changes in `AnacondaRecipes/arrow-cpp` that would be worthwhile here. It would probably better be if @msarahan could tell us what we need to improve upstream so Anaconda could get rid of them in their recipes.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -14,11 +14,13 @@ source:
     - 0001-ARROW-3711-C-Don-t-pass-CXX_FLAGS-to-C_FLAGS.patch
 
 build:
-  number: 1002
+  number: 1003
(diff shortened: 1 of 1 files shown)


mariusvniekerk:
Thanks for this one

PR #97, opened 2019-09-07
Move to Azure CI

xhochy:
See https://github.com/conda-forge/conda-forge.github.io/issues/857


recipe/bld.bat
```diff
@@ -1,7 +1,7 @@
 mkdir "%SRC_DIR%"\cpp\build
 pushd "%SRC_DIR%"\cpp\build
 
-cmake -G "%CMAKE_GENERATOR%" ^
+cmake -G "NMake Makefiles" ^
```
isuruf: I suggest using `jom` or `ninja` to make the build parallel.

xhochy: Is `ninja` working nowadays? We had also issues with it in Azure before...

isuruf: Yes. When did you have issues? I fixed some issues with Ninja several months ago.

xhochy: No longer any issues, thanks!

kszucs:
@xhochy nice! We need to back port it to the nightlies.

PR #101, opened 2019-09-25
Enable bz2 compression

xhochy:
Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -12,7 +12,7 @@ source:
   sha256: {{ checksum }}
 
 build:
-  number: 3
+  number: 4
   skip: true  # [win and vc<14]
   run_exports:
     - {{ pin_subpackage("arrow-cpp", max_pin="x.x") }}
@@ -27,6 +27,7 @@ requirements:
   host:
     - boost-cpp
     - brotli
+    - bzip2
     - c-ares
     - double-conversion
     - flatbuffers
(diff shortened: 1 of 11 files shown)


recipe/meta.yaml
```diff
@@ -27,6 +27,7 @@ requirements:
   host:
     - boost-cpp
     - brotli
+    - bzip2
```
pitrou: Does it need to be added to the "run" dependencies as well? I forget whether we link statically by default...

xhochy: No, the `bzip2` package has a `run_exports` section that defines what goes into `run` of the package that uses it in the `host` section. To check this, see:

```
% conda render recipe | grep 'run:' -A 20 | grep bzip2
        - bzip2 >=1.0.8,<2.0a0
```

pitrou: Ah, is this a new conda thing?
But if we were to link statically, it would still appear as a runtime dependency?

xhochy: Yes, I guess it came with 4.6. 

If we link statically, it will still appear but we should link dynamically on conda-forge.

xhochy: Build output shows we're linking dynamically:

```
   INFO (arrow-cpp,lib/libarrow.so.14.1.0): Needed DSO lib/libbz2.so.1.0 found in conda-forge::bzip2-1.0.8-h516909a_1
```

xhochy:
@pitrou Ok to merge this?

PR #123, opened 2020-03-15
Enable CUDA support.

pearu:
Checklist


pearu:

pearu:
There are CI failures:
```
2020-03-15T20:31:40.8273787Z Traceback (most recent call last):
2020-03-15T20:31:40.8277197Z   File "/opt/conda/lib/python3.7/site-packages/conda_build/environ.py", line 764, in get_install_actions
<snip>
2020-03-15T20:31:40.8292864Z     raise ResolvePackageNotFound(bad_deps)
2020-03-15T20:31:40.8293255Z conda.exceptions.ResolvePackageNotFound: 
2020-03-15T20:31:40.8293952Z   - nvcc_linux-64=None
```
What is the proper way to fix it?

pearu:
@isuruf do you have ideas how to fix the nvcc_linux-64=None issue reported above?

isuruf:
That's there to enable a build without cuda. Do you want a build without cuda? Or do you want only builds with cuda? If the former, use `- {{ compiler('cuda') }}   # [linux64 and cuda_compiler_version != 'None']. If the latter, use `skip: True  # [linux64 and cuda_compiler_version=='None']`

pearu:
Thanks @isuruf for the hints! The answer to your question actually depends on whether a single feedstock can support both cuda and non-cuda versions of the packages for the same platform (linux64 here) and how. The aim is that arrow-cpp would be installable to environments without cuda (the current master behavior) and to environments with cuda (the extension from this PR, the package needs to include `libarrow_cuda` library).

Btw, in this particular case of arrow-cpp package, any cuda version of the package would likely work in environments with no cuda installed, libarrow_cuda library would just not be unused.

jakirkham:
Would take a look at [`ucx-split`]( https://github.com/conda-forge/ucx-split-feedstock/blob/master/recipe/meta.yaml ) as a simple example of supporting CUDA and non-CUDA builds in the same feedstock.

pearu:
Thanks @jakirkham for the link! The current PR is using the same approach as ucx-split.

pearu:
Please review.
@xhochy

jakirkham:
Will we need similar changes for `pyarrow`?

isuruf:
Looks like this only links to `libcuda.so`. Then you can use 9.2 to build against and have one variant right?

jakirkham:
Not necessarily. It depends on what CUDA symbols they are using.

isuruf:
Can you expand on that? Looks to me like arrow is using the libcuda API and doesn't depend on new versions

isuruf:
See also the discussion in https://github.com/conda-forge/openmpi-feedstock/pull/54

kkraus14:
> Can you expand on that? Looks to me like arrow is using the libcuda API and doesn't depend on new versions

> See also the discussion in conda-forge/openmpi-feedstock#54


I'm almost 100% sure Arrow is using `libcudart` (runtime provided by toolkit) versus `libcuda` (driver). OpenMPI is an entirely different thing here.

isuruf:
> I'm almost 100% sure Arrow is using libcudart

If you look at the log, you'll see that it is not. (Or it is linking statically)

kkraus14:
> > I'm almost 100% sure Arrow is using libcudart
> 
> If you look at the log, you'll see that it is not. (Or it is linking statically)

From a quick glance:

```
2020-03-17T18:08:25.8329629Z -- CMAKE_C_FLAGS: -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/arrow-cpp-0.16.0 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -I/usr/local/cuda/include -O3 -DNDEBUG   -Wall
2020-03-17T18:08:25.8331866Z -- CMAKE_CXX_FLAGS:  -Wno-noexcept-type -fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/arrow-cpp-0.16.0 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -I/usr/local/cuda/include -fdiagnostics-color=always -O3 -DNDEBUG  -Wall 
2020-03-17T18:08:25.8901406Z -- Found CUDA: /usr/local/cuda (found version "10.1") 
2020-03-17T18:08:25.8916653Z -- CUDA Libraries: /usr/local/cuda/lib64/libcudart_static.a;-pthread;dl;$BUILD_PREFIX/x86_64-conda_cos6-linux-gnu/sysroot/usr/lib/librt.so
```
That's the runtime library, not the driver library but it is static linking it.

I do also see `libcuda.so` being linked and arrow is adding it in their CMake (https://github.com/apache/arrow/blob/master/cpp/src/arrow/gpu/CMakeLists.txt#L39) via the `${CUDA_CUDA_LIBRARY}`, but I'm not clear on why as from what I can see in their GPU code it looks to only be using the runtime API.

isuruf:
> but I'm not clear on why as from what I can see in their GPU code it looks to only be using the runtime API.

Can you point to the code you are referring to? I only looked at https://github.com/apache/arrow/blob/master/cpp/src/arrow/gpu/cuda_context.cc and all of the calls I looked at uses libcuda.so.

kkraus14:
> > but I'm not clear on why as from what I can see in their GPU code it looks to only be using the runtime API.
> 
> Can you point to the code you are referring to? I only looked at https://github.com/apache/arrow/blob/master/cpp/src/arrow/gpu/cuda_context.cc and all of the calls I looked at uses libcuda.so.

I was mistaken, apologies. Typically people would use the runtime API for these types of things.

isuruf:
No worries. Thanks for your input.

jakirkham:
Yeah sorry Isuru. I may not be able to answer your question as I don't know enough about Arrow's code base. Hopefully someone more knowledgeable there can provide us some idea of what CUDA functions they are using 🙂

xhochy:
> Will we need similar changes for `pyarrow`?

Yes, there is a `pyarrow._cuda` module which gets built only when CUDA is activated.

isuruf:
@xhochy, this PR should not have been merged. As it is right now, some users who don't even have a GPU will get a proprietary binary blob (cuda-toolkit).

xhochy:
> @xhochy, this PR should not have been merged. As it is right now, some users who don't even have a GPU will get a proprietary binary blob (cuda-toolkit).

Oh, then I didn't understand the how selection mechanism for cuda vs non-cuda works.

xhochy:
@isuruf Is there something easy to fix or should we revert and mark as broken?

isuruf:
It's better to mark as broken for now. There are some issues that needs to be discussed before fixing.

xhochy:
@pearu I revert the PR on master, please reopen and then we can discuss the issues @isuruf is talking about. 

@isuruf Please request changes in future, so I don't merge this accidentially again ;). From the discussion I thought everything was resolved.

jakirkham:
Sorry I'm confused @isuruf, what is the issue here?

isuruf:
Question to @xhochy. If arrow is built with CUDA enabled, do downstream packages need to enable CUDA too? (I know that downstream package can enable CUDA, but do they need to?)
If not, then we can have one single package with CUDA enabled and people who don't want CUDA will only have a small amount of code added.

@jakirkham, there's no features added here, so the conda solver is free to choose a CUDA build which brings in `cudatoolkit` (unless `ignore_run_exports` is used) which is a huge proprietary binary blob.

xhochy:
> Question to @xhochy. If arrow is built with CUDA enabled, do downstream packages need to enable CUDA too? (I know that downstream package can enable CUDA, but do they need to?)
> If not, then we can have one single package with CUDA enabled and people who don't want CUDA will only have a small amount of code added.

No, downstream packages without cuda will just work fine. The CUDA support/usage is contained in `libarrow_cuda${SHLIB_EXT}`. All other shared objects don't deal link/operate with CUDA.

jakirkham:
> @jakirkham, there's no features added here, so the conda solver is free to choose a CUDA build which brings in `cudatoolkit` (unless `ignore_run_exports` is used) which is a huge proprietary binary blob.

Got it. Yeah this seems like a feature we would want for other reasons. Missed that wasn't included here.

For context @pearu @xhochy, we handle this in the `ucx` case by adding this [`ucx-proc` output]( https://github.com/conda-forge/ucx-split-feedstock/blob/84dbdb4d4d58b7cbbe23f15803490658f95f7470/recipe/meta.yaml#L20-L32 ). This allows user selection of CUDA support or not.

Another way to solve this is to add some `dlopen` logic around loading the CUDA libraries you have built here with fallback handling if `libcuda` is not available. This is how `openmpi` solves this problem (as Isuru alluded to earlier).

jakirkham:
Sorry I may have misunderstood your comment before, @xhochy. Are you already using `dlopen`? If so, maybe just adding `cudatoolkit` to `ignore_run_exports` and `run_constrained` would solve this issue.

xhochy:
No, we're not using `dlopen`. I looked again a bit through the Arrow source code and the following happens:

* We build `libarrow_cuda` which links against "cuda", all other `libarrow_*` libraries are exactly the same, no changed functionality, no link to "cuda"
* We build the plasma store and `libplasma` differently based on whether we have CUDA support. Thus we need to have different Arrow builds for with-cuda and without-cuda.

xhochy:
Marking the builds as broken: https://github.com/conda-forge/cf-mark-broken/pull/22

jakirkham:
Ok then borrowing `ucx-proc` should work here (`arrow-proc`?).

pearu:
I am not sure how the splitting of arrow-cpp should work.

ucx-split seems to produce two conda packages:
- ucx - main package
- ucx-proc with cpu and gpu build labels - I am not sure what it does ?

I can think of the following solution:
- arrow-cpp - contains everything that the current master provides.
- arrow-cpp-cuda - contains arrow-cpp plus libarrow_cuda library (and cuda specific libplasma).
Can this be achieved within one feedstock?

Btw, I have created https://github.com/conda-forge/arrow-cpp-feedstock/pull/125 but it might be nonsense.

isuruf:
Here's an option. Have 2 variants.

1. With CUDA but with `__cuda >=9.2` as a run requirement. This package will be installed only if conda>=4.8 and the host has CUDA driver built
2. Without CUDA

jakirkham:
Though users may want to use the CPU-only package on machines that have a GPU.

isuruf:
@jakirkham, then they can use the build string to get the package that doesn't use the GPU.

jakirkham:
Build strings tend to have other things like build numbers, which can make this a bit hard to use in practice.

PR #125, opened 2020-03-19
Enable CUDA support [READY]

pearu:
Checklist


isuruf:
@pearu, your idea is a good one, but this is how you should implement it.
Skip `linux64 and cuda_compiler_version != 9.2`. (That means you only build with cuda 9.2 on linux64.) Have two packages `arrow-cpp` and `arrow-cpp-cuda`.

pearu:
@isuruf ok, I'll fix to using 9.2. Two questions:
1. How to fix the following osx CI failure:
```
Error: Failed to render jinja template in /Users/runner/runners/2.165.2/work/1/s/recipe/meta.yaml:
'cuda_compiler_version' is undefined
```
2. When building arrow-cpp against cuda 9.2, will the conda package still be usable in environments with cuda > 9.2? Why not build against the latest cuda version?

isuruf:
> When building arrow-cpp against cuda 9.2, will the conda package still be usable in environments with cuda > 9.2? Why not build against the latest cuda version?

Yes, it looks like arrow is using only the CUDA driver API. @xhochy should confirm.

pearu:

recipe/meta.yaml
```diff
@@ -1,119 +1,138 @@
...
-    - orc  # [unix]
-    - python
-    - rapidjson
-    - re2
-    - snappy
-    - thrift-cpp >=0.11
-    - zlib
-    - zstd
+outputs:
+  - name: arrow-cpp{{ cuda_ext }}
```
jakirkham: Historically we have made a separate mutex metapackage for this kind of selection. That way the user can choose what this should be pretty easily. Also it is less dependent on the solver this way.

jakirkham: This mutex metapackage can also be useful for [aligning other downstream packages]( https://github.com/conda-forge/ucx-split-feedstock/blob/84dbdb4d4d58b7cbbe23f15803490658f95f7470/recipe/meta.yaml#L54-L55 ).

isuruf: I think https://github.com/conda-forge/arrow-cpp-feedstock/pull/123#issuecomment-601320569 is a better solution.

isuruf: Nvm, `libplasma` is built differently.

pearu: @jakirkham, the `ucx` example is confusing for me as it is not clear where the cuda-specific build results are located: the `ucx-proc` packages (builds `cpu` and `gpu`) are empty and there is no build string specified for the `ucx` package whereas the packages in  https://anaconda.org/conda-forge/ucx/files seem to be cuda specific only.

For arrow-cpp, we need to support at least two scenarios:
1. using arrow-cpp from cpu-only specific software.
2. using arrow-cpp from software that can take advantage of cuda supported gpus. One can assume that the software is linked to cuda libraries.

So, there must exist at least two separate conda builds of arrow-cpp, one build with `-DARROW_CUDA=OFF` and one with `-DARROW_CUDA=ON`. Can you explain (or direct me to relevant documentation or examples) how do you foresee the above to be achievable by using mutex metapackages? As I understand, metapackages are useful for constraining solvers to specific package builds but the package builds must also exist somehow.

Another confusing matter is the choice of build strings: `gpu` is often specified to mean `cuda` but there exist `cuda` alternatives (`rocm`, for instance). While mutex metapackage works for cases with two mutually exclusive choices (`nocuda/cuda`, `mkl/nomkl`, etc), will it work for cases with more than two exclusive choices like `cuda/rocm/no-cuda-no-rocm`? Just asking out-of-curiosity...

xhochy: Is in the `ucx` version somehow a trick to ensure that if `gpu` is not requested, only the `cpu` package is installed? Triggering `conda` to prefer one version over the other is making this hard to approach for me.

jakirkham: @xhochy, currently `ucx` doesn't do that, but it is possible to do that ( https://github.com/conda-forge/ucx-split-feedstock/issues/26 ).

@pearu, the `ucx` package hash changes based on whether CUDA was used and what version of CUDA was used. `ucx-proc` and `cudatoolkit` provide the constraints to force a `cpu` or `gpu` build along with CUDA version constraints.

pearu: Thanks, @jakirkham. Is there any harm in using build strings on top of hash changes?

jakirkham: The main issue with changing the `build/string` field is this replaces the whole build string (as opposed to appending it or something else less drastic). The result is it hides the build number, package hash, Python information, and anything else Conda-build would have added to the build string automatically. It's possible to include this in the manually constructed build string. However it's a bit fragile if Conda-build changes the content in build strings later. So mostly I try to avoid this in non-metapackages. Though completely up to you whether you pursue that approach. 🙂

jakirkham: FWIW it is possible to inspect the package hash to see what is contained `conda inspect hash-inputs`.

cc @msarahan (who convinced me at one point to just use package hashes 😉)

recipe/meta.yaml
```diff
@@ -1,123 +1,159 @@
...
-    - snappy
-    - zlib
-    - zstd
+  - name: arrow-cpp
+    script: build-arrow.sh  # [not win]
+    script: bld.bat  # [win]
+    version: {{ version }}
+    build:
+      number: {{ number }}
+      string: "{{ build_ext }}"
```
isuruf:
```suggestion
      string: "{{ number }}_h{{ PKG_HASH }}_{{ build_ext }}"
```

recipe/meta.yaml
```diff
@@ -1,123 +1,159 @@
...
+      skip: true  # [(win and vc<14) or (linux and cuda_compiler_version is not undefined and cuda_compiler_version != '9.2' and cuda_compiler_version != 'None')]
```
isuruf: Since you are building for the oldest, you need to add,
```
ignore_run_exports:
  - cudatoolkit
```

recipe/meta.yaml
pearu: ok, gotcha.
To complete the story, how the users are able to select which package to install?
When the build string contains an only hash, will this work:
```
conda install "arrow-cpp=*=*cuda"
```
or should it be `conda install arrow-cpp arrow-cpp-build-ext=*=cuda` or smth different?

recipe/meta.yaml
pearu: I guess it would be good to keep the python version in the build string as well.

recipe/meta.yaml
jakirkham: It would be the latter. They may additionally do the following to constrain the CUDA version.

```
conda install arrow-cpp arrow-cpp-build-ext=*=cuda cudatoolkit=10.1
```

isuruf: `conda install "arrow-cpp=*=*cuda"` will work too

jakirkham: It sounds like that may be removed from the build string, in which case it wouldn't.

recipe/meta.yaml
pearu: Fixed.

recipe/meta.yaml
pearu: Using
```
string: "py{{ CONDA_PY }}h{{ PKG_HASH }}_{{ number }}_{{ build_ext }}"
```
that matches with the default build string plus the build_ext part.

jakirkham:
@kkraus14, would you be able to take a look as well? 🙂

recipe/meta.yaml
```diff
@@ -1,123 +1,161 @@
...
+        - glog
+        - grpc-cpp
+        - lz4-c
+        - python
+        - re2
+        - snappy
+        - zlib
+        - zstd
+      run_constrained:
+        - arrow-cpp-build-ext * {{ build_ext }}
```
isuruf:
```suggestion
        - arrow-cpp-build-ext * {{ build_ext }}
        - cudatoolkit >=9.2
```

pearu:
@jakirkham @isuruf @xhochy thank you all for your feedback on this PR!

In sum, to install arrow-cpp with cuda enabled, one should use the following commands:
```
conda install arrow-cpp arrow-cpp-build-ext=*=cuda cudatoolkit=<CUDA version>
conda install "arrow-cpp=*=*cuda"
```
I presume that the later is able to pick up the CUDA version from environment correctly using `CUDA_HOME` or the location of `cuda-gdb`.

Please review.

recipe/meta.yaml
pearu: Fixed, thanks!

pearu:
> You still need to add
> 
> ```
> build:
>   track_features:
>     - arrow-cuda
> ```
> 
> for the cuda builds in `arrow-cpp` to make sure the gpu builds are weighed down.
> 

Done.

Can the name of the feature be arbitrary like `arrow-cuda` currently is?

pearu:
DO NOT MERGE: there is a bug that disables cuda build for cuda enabled arrow-cpp.

jakirkham:
What's the bug? 🙂

pearu:
Notice that CUDA extension is disabled in:
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=135074&view=logs&j=4f922444-fdfe-5dcf-b824-02f86439ef14&t=937c195f-508d-5135-dc9f-d4b5730df0f7&l=822

There was a bug in build-arrow.sh script where the `if` block was broken so that the `else` part was always effective. The last commit fixes it.

The next issue is that for some reason CUDA_HOME is not defined. I suspect that `compiler("cuda")` is not effective (locally I don't see nvcc in build deps).

pearu:
The issue looks very similar to https://github.com/conda-forge/nvcc-feedstock/issues/30 .

pearu:
When building the recipe locally, I get
```
The following specifications were found to be incompatible with your CUDA driver:

  - feature:/linux-64::__cuda==10.2=0
  - feature:|@/linux-64::__cuda==10.2=0

Your installed CUDA driver is: 10.2

Note that strict channel priority may have removed packages required for satisfiability.
```
Yes, locally I have installed CUDA driver version 10.2 but I have also cudatoolkit 9.2 installed and made effective via `CUDA_HOME (=/usr/local/cuda-9.2.148)`.

Why conda build uses CUDA driver version? Isn't that irrelevant for building cuda enabled software?

jakirkham:
Sorry it's quite late for me. So I'll try to look tomorrow.

That said, maybe PR ( https://github.com/conda/conda/pull/8267 ) would provide some context as to what Conda is doing.

pearu:
Defining `CONDA_OVERRIDE_CUDA="9.2"` and disabling `compiler("cuda")` makes local conda build successful (hence my last suspicious commit).

jakirkham:
Though it’s now not building with CUDA in the feedstock

jakirkham:
If you want to build locally, would just run the `build-locally.py` script (note it needs `shyaml` installed to parse the configuration files)

pearu:
> Though it’s now not building with CUDA in the feedstock

Actually it did build with CUDA because I did not rerender.

The next step is to figure out how to enable `compiler("cuda")` and resolve
```
conda.exceptions.ResolvePackageNotFound: 
  - cuda_linux-64=9.2
```
in the CI.

pearu:
The cuda enabled build issues (bug in bash script, using cuda_enabled in constraints, need to rerender) are now resolved. @jakirkham @isuruf @xhochy please rei

isuruf:
So, the two ways to install arrow-cpp with cuda are,

1. `conda install arrow-cpp arrow-cpp-build-ext=*=cuda`
2. `conda install "arrow-cpp=*=*cuda"`

I'm not sure why 1 is better than 2. 1 unnecessarily creates a `arrow-cpp-build-ext` package.

pearu:
I don't have a definite answer and I wonder that myself but my guess is that `arrow-cpp-build-ext` can be useful if other packages can also rely on it. For instance, `pyarrow`. If it makes sense then perhaps such a meta-package should be defined in a more general setting where `arrow-cpp`, `pyarrow`, and other packages possibly using CUDA could use it.

jakirkham:
Did you specify the variant from `.ci_support` when building locally? That error suggests it wasn’t specified.

Anyways the point of `build-locally.py` and the Docker image is to provide a consistent build environment so that we don’t need to debug users’ systems and configurations. So if possible, please just use that for building locally.

isuruf:
> If it makes sense then perhaps such a meta-package should be defined in a more general setting where arrow-cpp, pyarrow, and other packages possibly using CUDA could use it.

In `pyarrow` case `pyarrow=*=*cuda` needs to have a dependency on `arrow-cpp=*=*cuda`. So I'm not sure that `arrow-cpp-build-ext` is needed even then.

pearu:
> Did you specify the variant from `.ci_support` when building locally? That error suggests it wasn’t specified.

No, I am using local conda_build_config.yaml . But I will look into using build-locally.py.

The current problem with this PR seems to be the following:
- for selecting cuda enabled extension of arrow-cpp, should we use mutex metapackage approach or build string approach?

The PR currently supports both approaches. I would be fine to leave it as it is (in order to move on) as well as dropping one of the approaches (to minimize user choices to a preferred one, whichever it is).

pearu:
@jakirkham @isuruf do you have any recommendations regarding the problem stated above?

isuruf:
My recommendation is build string. https://github.com/conda-forge/arrow-cpp-feedstock/pull/125#issuecomment-603245704

jakirkham:
My recommendation would a build string on a mutex package.

isuruf:
@jakirkham, can you explain why option 1 is better than option 2 at https://github.com/conda-forge/arrow-cpp-feedstock/pull/125#issuecomment-603245704? Adding more packages than necessary is confusing.

jakirkham:
Sure. First using a package that is not dependent on means the solver doesn't have to consider other things when trying to respect this constraint (so less likely to get a broken or unintended solve). Second manipulating the package string on ordinary packages has the potential to lose useful information (build number, build hash, etc.) that are added to the build string.

isuruf:
> First using a package that is not dependent on means the solver doesn't have to consider other things when trying to respect this constraint (so less likely to get a broken or unintended solve).

I don't understand this part. Can you explain more? I don't see how an unintended solve is going to occur if `arrow-cpp=*=*cuda` is used instead of `arrow-cpp arrow-cpp-build-ext=*=cuda`.

> Second manipulating the package string on ordinary packages has the potential to lose useful information (build number, build hash, etc.) that are added to the build string.

This is not relevant as the build string here is added correctly.

jakirkham:
What happens if the user specifies `arrow-cpp` with a spec? What happens if they run `conda update --all`? What happens if the user updates something else in the environment either a direct dependency (like `python`) or something else?

I disagree. The problem is that we are clobbering the build string. There is no clean way to append to it. As soon as some other information gets added by conda-build, this will be missed and need to be added manually once we figure out what was missed.

pearu:
> I disagree. The problem is that we are clobbering the build string. There is no clean way to append to it. As soon as some other information gets added by conda-build, this will be missed and need to be added manually once we figure out what was missed.

The current build string here is an extension of the default one:
```
string: "py{{ CONDA_PY }}h{{ PKG_HASH }}_{{ number }}_{{ build_ext }}"
```
and, IIUC, the `PKG_HASH` part should handle any future information additions without the need to update the build string manually. Is your concern about the default build string changing in the future?

jakirkham:
That or changes in the dependencies here that would otherwise change the default build string, but won't in our case without manual modification.

Though there is a third situation that I ran into and forgot to mention. It seems difficult to work with partial build strings in Conda-build recipes in practice. Here's [an example of one issue]( https://github.com/conda-forge/conda-forge.github.io/issues/867#issuecomment-595006878 ) that can be run into here.

scopatz:
Hi @jakirkham @isuruf - I have a couple of questions here.

* How are we specifying the cuda dependence for other packages that have a CUDA option (like pytorch?)
* It seems like we can always change which way we are doing things if we want to in the future. `arrow-cpp=*=*cuda` seems cleaner to the user to me, but again, it doesn't really matter.

I feel like this shouldn't be blocked on having a broader discussion of how we want to do this generally.

jakirkham:
This doesn't seem like the right place for that discussion @scopatz. This PR is just trying to provide CUDA build support at staged-recipes. Can we please move that back to the arrow feedstock PR?

jakirkham:
Other packages are using a mutex package.

isuruf:
> It seems like we can always change which way we are doing things if we want to in the future. arrow-cpp=*=*cuda seems cleaner to the user to me, but again, it doesn't really matter.

It's hard to change stuff in the future. Get it right first, otherwise people will complain.

That said, we can have both. We can have the build string in it and users who want to use the mutex package can do so.

We should also have a look at the name. `arrow-cpp-build-ext`, `ucx-proc`, etc.

pearu:
I have fixed merged conflicts and renamed `arrow-cpp-build-ext` to `arrow-cpp-proc`.

@isuruf @jakirkham please review.

recipe/meta.yaml
```diff
@@ -1,124 +1,166 @@
...
-    - zstd
+  - name: arrow-cpp
+    script: build-arrow.sh  # [not win]
+    script: bld.bat         # [win]
+    version: {{ version }}
+    build:
+      number: {{ number }}
+      string: "py{{ CONDA_PY }}h{{ PKG_HASH }}_{{ number }}_{{ build_ext }}"
+      skip: true  # [win and vc<14]
+      skip: true  # [linux and cuda_compiler_version is not undefined and cuda_compiler_version != '9.2' and cuda_compiler_version != 'None']
```
jakirkham, line 47: We might be able to simplify this like so. IOW other platforms where CUDA is not yet supported (`undefined`), CPU-only builds on platforms that do support CUDA (`None`), and CUDA 9.2 are permitted. Otherwise we skip all other CUDA versions. Please let me know if that matches the intent or if I'm missing something here.

```suggestion
      skip: true  # [cuda_compiler_version not in (undefined, "None", "9.2")]
```

recipe/meta.yaml
```diff
@@ -1,124 +1,166 @@
...
+      ignore_run_exports:
+        - cudatoolkit
+      track_features:
+        - arrow-cuda         # [linux and cuda_compiler_version is not undefined and cuda_compiler_version == '9.2']
```
jakirkham, line 51: Should be able to just check the CUDA version for what we need here. The fact that Linux is the only platform supported may not always be the case in the future. This will make it easier to adopt those platforms as they become available.

```suggestion
      track_features:
        - arrow-cuda         # [cuda_compiler_version is not undefined and cuda_compiler_version == '9.2']
```

Alternatively we could use Jinja here. It may be possible to drop the `else`  as well.

```suggestion
      track_features:
        {{ "- arrow-cuda" if cuda_enabled else "" }}
```

recipe/meta.yaml
```diff
@@ -1,124 +1,166 @@
...
+    requirements:
+      build:
+        - cmake
+        - autoconf  # [unix]
+        - ninja
+        - make      # [unix]
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - {{ compiler("cuda") }}  # [linux and cuda_compiler_version is not undefined and cuda_compiler_version == '9.2']
```
jakirkham, line 60: Since this already uses Jinja, we might consider condensing this all into Jinja.

```suggestion
        {{ ("-" ~ compiler("cuda")) if cuda_enabled else "" }}
```

recipe/meta.yaml
```diff
@@ -1,124 +1,166 @@
...
-    - if exist %PREFIX%\\Library\\lib\\arrow_python_static.lib exit 1   # [win]
-    - if exist %PREFIX%\\Library\\lib\\parquet_static.lib exit 1        # [win]
-    - if exist %PREFIX%\\Library\\lib\\gandiva_static.lib exit 1        # [win]
-
-    # conda tools
-    - conda inspect linkages -p $PREFIX $PKG_NAME  # [not win]
-    - conda inspect objects -p $PREFIX $PKG_NAME   # [osx]
+        # conda tools
+        - conda inspect linkages -p $PREFIX $PKG_NAME  # [not win]
+        - conda inspect objects -p $PREFIX $PKG_NAME   # [osx]
```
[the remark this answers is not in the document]
pearu: Fixed.

recipe/meta.yaml
pearu: The suggested skip condition is as intended, so let's use it.

recipe/meta.yaml
pearu: Fixed.

recipe/meta.yaml
pearu: Fixed.

recipe/meta.yaml
```diff
@@ -1,124 +1,164 @@
...
+        {{ "- arrow-cuda" if cuda_enabled else "" }}
+    requirements:
+      build:
+        - cmake
+        - autoconf  # [unix]
+        - ninja
+        - make      # [unix]
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        {{ ("- " ~ compiler("cuda")) if cuda_enabled else "" }}
```
isuruf, line 60: This is going to interfere with conda-smithy/conda-build's used variable detection

recipe/meta.yaml
isuruf: This is a bad idea. conda-build's variant detection is text based and this interferes with used variable detection as seen in https://github.com/conda-forge/arrow-cpp-feedstock/pull/125/commits/bf07c1a3d3ff9c6dc2e0fa743ea15233e729cb1e where `cuda_compiler` variable is removed.

recipe/meta.yaml
pearu: Does it mean that Jinja cannot be used here?

isuruf: Yes, you can't. You have to use `- {{ compiler("cuda") }}` with a selector without jinja.

pearu: Fixed.

pearu:
Hi all, could we have this PR merged?

pearu:
Several projects using CUDA rely on arrow-cpp 0.16 and packaging those are currently blocked by this PR.
@xhochy could you merge this PR?

xhochy:
@pearu Leaving this open until tomorrow to let @isuruf give any objections, otherwise I'm going to have a review by myself then (no time for that today). I will merge this to a separated 0.16 branch and will port things over to 0.17 afterwards.

recipe/meta.yaml
jakirkham: Hmm...that's unfortunate. Thanks for catching though Isuru. Was trying to find a way to simplify the selection logic, but agree conda-build makes that a bit tricky today.

jakirkham:
LGTM. Thanks for your work on this @pearu! 😄

alexbaden:
Hi all -- thanks for doing this work @pearu and to everyone else for reviewing. We just pushed out a new version of OmniSciDB which uses Arrow 0.16 and are interested in seeing this patch merged so we can push our python connector changes as well. Is there a rough ETA?

jakirkham:
FWIW this issue may be relevant.

xref: https://issues.apache.org/jira/browse/ARROW-8608

PR #137, opened 2020-05-14
Add cuda support for master

xhochy:
Checklist


xhochy:
@pearu I took a stab at copying over the CUDA changes to 0.17/master, can you take a look?

recipe/meta.yaml
```diff
@@ -1,120 +1,158 @@
...
+      track_features:
+        {{ "- arrow-cuda" if cuda_enabled else "" }}
+    requirements:
+      build:
+        - cmake 3.16.*
+        - autoconf  # [unix]
+        - ninja
+        - make  # [unix]
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
```
pearu, line 56: I think the following dependency is required in build:
```
    - {{ compiler("cuda") }}  # [cuda_compiler_version is not undefined and cuda_compiler_version
 == '9.2']
```

xhochy:
@conda-forge-admin please rerender

pearu:
Re failing win builds: the build log shows
```
2020-05-15T07:53:41.8464152Z Traceback (most recent call last):
2020-05-15T07:53:41.8464746Z   File "C:\Miniconda\lib\site-packages\conda\gateways\disk\create.py", line 245, in make_menu
2020-05-15T07:53:41.8465293Z     import menuinst
2020-05-15T07:53:41.8465780Z   File "C:\Miniconda\lib\site-packages\menuinst\__init__.py", line 23, in <module>
2020-05-15T07:53:41.8466306Z     from .win32 import Menu, ShortCut
2020-05-15T07:53:41.8466796Z   File "C:\Miniconda\lib\site-packages\menuinst\win32.py", line 19, in <module>
2020-05-15T07:53:41.8467341Z     from .winshortcut import create_shortcut
2020-05-15T07:53:41.8474996Z ModuleNotFoundError: No module named 'menuinst.winshortcut'done
```
Following https://github.com/conda/conda/issues/1762#issuecomment-153366413, perhaps try adding  `menuinst` to build dependencies?

Another sign of failure is
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=162703&view=logs&j=5be07ae1-d8ba-5406-47b6-8e3a3a12f825&t=ac199efb-6791-5d5e-770f-2831854227bc&l=12
for which I don't have interpretation. @isuruf ?

Finally, what is the reason for restricting to `cmake 3.16`?

xhochy:
I suspect that the build script being the default name `bld.bat` could have been the issue. Let's see what CI tells us.

The reason to pin to `cmake 3.16` is that Arrow doesn't detect the correct Python with `cmake 3.17`, this should be fixed again in `3.18`, then we can remove the constraint again.

xhochy:
Seems like renaming helped a lot. I will revert the `ppc64le` builds again to Travis to get them back to speed.

recipe/meta.yaml
```diff
@@ -1,120 +1,159 @@
...
-    - zlib
-    - zstd
+  - name: arrow-cpp
+    script: build-arrow.sh  # [not win]
+    script: bld-arrow.bat   # [win]
+    version: {{ version }}
+    build:
+      number: {{ number }}
+      string: "py{{ CONDA_PY }}h{{ PKG_HASH }}_{{ number }}_{{ build_ext }}"
+      skip: true  # [cuda_compiler_version not in (undefined, "None", "9.2")]
```
h-vetinari, line 44:
```suggestion
```

I think it's worth trying to see whether cuda 10.x builds successfully. 9.2. cannot make use of the `compute_75` instruction set, for example.

h-vetinari: Following this suggestion would need a few more adaptations (using `# [cuda_compiler_version is not undefined and cuda_compiler_version != "None"]` as a selector), plus a rerender

isuruf: That's not necessary. Please read the discussion in the PR that added CUDA support.

h-vetinari: OK, thanks for the info. To record the outcome (to the best of my understanding): since arrow only uses the cuda driver (and not the toolkit), building for one cuda version is compatible with all.

pearu:
LGTM.

Btw, are there tricks that could avoid the issue describe [here](https://github.com/conda-forge/pyarrow-feedstock/pull/101#issuecomment-629475808) that could be applied in this PR to avoid future problems?

xhochy:
> LGTM.
> 
> Btw, are there tricks that could avoid the issue describe [here](https://github.com/conda-forge/pyarrow-feedstock/pull/101#issuecomment-629475808) that could be applied in this PR to avoid future problems?

Yes, we should make better recipes :) @isuruf thankfully already fixed the major breakage and with https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/618 we will have a consistent CF global linkage with re2.

PR #146, opened 2020-05-20
Include CUDA enabled pyarrow

### LINKED ISSUE #93: Split package?
Should we split this package into a portion without python and one with the python bindings?

Feels a little weird that arrow-cpp requires a python to be installed at all times.

Thoughts?

pearu:
Fixes https://github.com/conda-forge/pyarrow-feedstock/issues/107
Fixes #93 

Checklist


jakirkham:
Thanks for picking this up @pearu! 😄

cc @h-vetinari (who also seemed interested in this)

h-vetinari:
Thanks for the ping @jakirkham; looks good @pearu!

Also closes #93 (see discussion there). Finally, merging this means the pyarrow-feedstock must be disabled/archived somehow - don't know how that works.

recipe/meta.yaml
```diff
@@ -165,6 +165,8 @@ outputs:
       number: {{ number }}
       string: "{{ build_string }}"
       skip: true  # [cuda_compiler_version not in (undefined, "None", "9.2")]
+      # Skip aarch64 as the build fails with: Unsupported arch flag: -march=.
+      skip: true  # [aarch64]
```
h-vetinari, line 169: AFAICS, the pyarrow feedstock builds aarch64 just fine, although the cuda-PR on `0.16.x` is building on azure instead of drone. In any case, we shouldn't be skipping so easily, IMO.

Can you try the following in `conda-forge.yml`?
```diff
 appveyor:
   secure: {BINSTAR_TOKEN: tumuXLL8PU75WMnRDemRy02ruEq2RpNxeK3dz0MjFssnosPm2v4EFjfNB4PTotA1}
-provider: {linux_aarch64: default, linux_ppc64le: default, win: azure}
+provider: {linux_aarch64: default, linux_ppc64le: azure, win: azure}
 travis:
```

pearu:
pyarrow fails to build on aarch64:
```
CMake Error at cmake_modules/SetupCxxFlags.cmake:338 (message):
  Unsupported arch flag: -march=.
  CMakeLists.txt:100 (include)
```
I have disabled aarch64 for pyarrow but this does not feel right.
Does anyone have an idea of how to fix the pyarrow build for aarch64?

recipe/meta.yaml
pearu: @h-vetinari, thanks for the hint! It seems that building on azure is working out better indeed.

pearu: It didn't work, so now using another workaround.

pearu:
@xhochy @h-vetinari @jakirkham Please review.

recipe/meta.yaml
```diff
@@ -150,19 +157,80 @@ outputs:
...
+        - make  # [unix]
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        # pyarrow does not require nvcc but it needs to link against libraries in arrow-cpp=*=*cuda
+        - {{ compiler("cuda") }}  # [cuda_compiler_version not in (undefined, "None")]
+      host:
+        # directly pin boost-cpp as we also seem to directly include boost symbols
+        # in the Python modules.
+        - arrow-cpp ={{ version }}=*{{ build_ext }}
+        - boost-cpp
```
h-vetinari, line 183: The comment for boost and the line where it's on don't fit together anymore.

h-vetinari: @pearu, in contrast to the pyarrow-feedstock, I'm not a collaborator here, so I cannot push to this PR's branch. In any case, there's only details left from my POV.

pearu: Thanks, @h-vetinari, I'll fix the nit and wait for an extra 24h to give anyone time for review.

pearu: Fixed.

pearu:
We have:
```
Cloning into '/tmp/tmp5w59ji6d_recipe/feedstock-outputs'...
validation results:
{
  "linux-64/pyarrow-0.17.1-py38h1234567_3_cpu.tar.bz2": false,
  "linux-64/arrow-cpp-proc-1.0.0-cpu.tar.bz2": true,
  "linux-64/arrow-cpp-0.17.1-py38h1234567_3_cpu.tar.bz2": true
}
NOTE: Any outputs marked as False are not allowed for this feedstock.
```
IIUC, the output of this feedstock cannot have the same name as the name of another (pyarrow) feedstock.

Is this analysis correct and what would be the proper procedure to move pyarrow under arrow-cpp?

h-vetinari:
```diff
-  - name: pyarrow
+  - name: arrow-py
```

I don't think that's the right approach. We want this feedstock to serve both packages, and have the packages not interfere with each other (which is automatic if they have the same name, otherwise we need another mutex etc.).

@conda-forge/core How can we absorb one feedstock into another correctly?

pearu:
I was about to propose https://github.com/conda-forge/parquet-cpp-feedstock approach where the future pyarrow packages would just point to arrow-py.
But I second the question of how to absorb feedstocks correctly.

xhochy:
We should be building `pyarrow` here. I have the superpowers to enable that but I'll have to consult @beckermr on how to use it ;)

recipe/build-pyarrow.sh
```diff
@@ -0,0 +1,46 @@
...
+export SETUPTOOLS_SCM_PRETEND_VERSION=$PKG_VERSION
+export PYARROW_BUILD_TYPE=release
+export PYARROW_WITH_DATASET=1
+export PYARROW_WITH_FLIGHT=1
+if [[ "$(uname -m)" = "ppc64le" || "$(uname -m)" = "aarch64" ]]
+then
+  export PYARROW_WITH_GANDIVA=0
+else
+  export PYARROW_WITH_GANDIVA=1
+fi
```
xhochy, line 18: We should build this unconditionally with 0.17
```suggestion
export PYARROW_WITH_GANDIVA=1
```

recipe/meta.yaml
```diff
@@ -150,19 +157,78 @@ outputs:
         - if exist %PREFIX%\\Library\\lib\\parquet_static.lib exit 1        # [win]
         - if exist %PREFIX%\\Library\\lib\\gandiva_static.lib exit 1        # [win]
 
+  - name: arrow-py
```
xhochy, line 160:
```suggestion
  - name: pyarrow
```

recipe/meta.yaml
```diff
@@ -150,19 +157,78 @@ outputs:
...
+      build:
+        - cmake 3.16.*
+        - ninja
+        - make  # [unix]
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        # pyarrow does not require nvcc but it needs to link against libraries in arrow-cpp=*=*cuda
+        - {{ compiler("cuda") }}  # [cuda_compiler_version not in (undefined, "None")]
+      host:
+        - arrow-cpp ={{ version }}=*{{ build_ext }}
```
xhochy, line 180: We should be able now to do `- {{ pin_subpackage('arrow-cpp', exact=True) }}`, can you check whether that works?

recipe/meta.yaml
```diff
@@ -150,19 +157,78 @@ outputs:
...
         - LICENSE.txt
-      summary: C++ libraries for Apache Arrow
+      summary: Python libraries for Apache Arrow
+
+    test:
+      imports:
+        - pyarrow
+        - pyarrow.dataset
+        - pyarrow.flight
+        - pyarrow.gandiva  # [not (aarch64 or ppc64le)]
```
xhochy, line 215:
```suggestion
        - pyarrow.gandiva
```

recipe/meta.yaml
pearu: Yes, it works (tested locally).

recipe/build-pyarrow.sh
pearu: Fixed.

recipe/meta.yaml
pearu: Fixed.

recipe/meta.yaml
pearu: Restored pyarrow as package name.

pearu:
@xhochy I have addressed all your comments. Surprisingly, there are no validation failures now.
Can you land this PR?

xhochy:
Requires https://github.com/conda-forge/feedstock-outputs/pull/3

There should have been validation failures 🤔 

I'll test this locally and merge then.

pearu:
The message
```
Cloning into '/tmp/tmpeilv4_v4_recipe/feedstock-outputs'...
validation results:
{
  "linux-64/arrow-cpp-0.17.1-py37h1234567_3_cuda.tar.bz2": true,
  "linux-64/pyarrow-0.17.1-py37h1234567_3_cuda.tar.bz2": false,
  "linux-64/arrow-cpp-proc-1.0.0-cuda.tar.bz2": true
}
NOTE: Any outputs marked as False are not allowed for this feedstock.
```
persists. I guess your superpowers will be needed after all, @xhochy .

xhochy:
@conda-forge-admin please restart ci

pearu:
Validate seems to be ok with the feedstock:
```
+ validate_recipe_outputs arrow-cpp-feedstock
Cloning into '/tmp/tmp95n9c6e__recipe/feedstock-outputs'...
validation results:
{
  "linux-64/arrow-cpp-0.17.1-py37h1234567_3_cuda.tar.bz2": true,
  "linux-64/pyarrow-0.17.1-py37h1234567_3_cuda.tar.bz2": true,
  "linux-64/arrow-cpp-proc-1.0.0-cuda.tar.bz2": true
}
NOTE: Any outputs marked as False are not allowed for this feedstock.
```

h-vetinari:
@pearu, can you put `Fixes #93` in the OP?

pearu:
> @pearu, can you put `Fixes #93` in the OP?

Done in the description of the issue.

h-vetinari:
Regarding the naming conventions, I think it *might* be worth renaming `arrow-cpp` to `libarrow`. This is in line with a lot of other feedstocks, but obviously a matter of taste. In this case, we'd have to have a compatibility output of `arrow-cpp` that depends on `libarrow`, which could be dropped after a few versions.

@isuruf made me aware of this in the context of [faiss](https://github.com/conda-forge/faiss-split-feedstock/blob/master/recipe/meta.yaml). Other examples I can think of off the top of my head are [blas](https://github.com/conda-forge/blas-feedstock/blob/master/recipe/meta.yaml) & [lapack](https://github.com/conda-forge/lapack-feedstock/blob/master/recipe/meta.yaml), [openblas](https://github.com/conda-forge/openblas-feedstock/blob/master/recipe/meta.yaml), [opencv](https://github.com/conda-forge/opencv-feedstock/blob/master/recipe/meta.yaml), [postgresql](https://github.com/conda-forge/postgresql-feedstock/blob/master/recipe/meta.yaml), [gdal](https://github.com/conda-forge/gdal-feedstock/blob/master/recipe/meta.yaml), plus a [whole bunch](https://github.com/conda-forge?q=lib) more (non-exhaustive).

Of course, this recipe has some counter examples to this: [aws-sdk-cpp](https://github.com/conda-forge/aws-sdk-cpp-feedstock/blob/master/recipe/meta.yaml), [boost-cpp](https://github.com/conda-forge/boost-cpp-feedstock/blob/master/recipe/meta.yaml), [grpc-cpp](https://github.com/conda-forge/grpc-cpp-feedstock/blob/master/recipe/meta.yaml), [thrift-cpp](https://github.com/conda-forge/thrift-cpp-feedstock/blob/master/recipe/meta.yaml), but I think they are far in the [minority](https://anaconda.org/search?q=-cpp) and the same argument could be made for renaming those (maybe worth noting that at least the last two seem to have been started by the arrow team).

Happy to open an issue for this if this needs more discussion.

isuruf:
`-cpp` was a trend that I started with `boost-cpp`. That was a mistake. I'm in favour of changing it, but this is not the correct place to raise the issue.

recipe/build-pyarrow.sh
```diff
@@ -0,0 +1,41 @@
...
+then
+    export PYARROW_CMAKE_OPTIONS="-DARROW_ARMV8_ARCH=armv8-a"
+fi
+
+cd python
+
+$PYTHON setup.py \
+        build_ext $BUILD_EXT_FLAGS \
+        install --single-version-externally-managed \
+                --record=record.txt
```
jakirkham, line 41: Are we able to use `pip` here?

recipe/bld-pyarrow.bat
```diff
@@ -0,0 +1,31 @@
...
+SET PYARROW_WITH_DATASET=1
+SET PYARROW_WITH_FLIGHT=1
+SET PYARROW_WITH_GANDIVA=1
+SET PYARROW_WITH_PARQUET=1
+SET PYARROW_CMAKE_GENERATOR=Ninja
+
+%PYTHON%   setup.py ^
+           build_ext ^
+           install --single-version-externally-managed ^
+                   --record=record.txt
```
jakirkham, line 29: Same question about `pip` here

xhochy:
> Happy to open an issue for this if this needs more discussion.

Please do so at https://github.com/conda-forge/conda-forge.github.io

recipe/build-pyarrow.sh
xhochy: We should check that but in a separate PR.

xhochy:
Thanks @pearu , let's 🙏 this works!

recipe/build-pyarrow.sh
jakirkham: Yeah that makes sense. Was having a little trouble seeing how this was built before. If this is no different from what we were already doing, agree we don't need to change it here.

pearu: Using pip does not work out of the box:
```
$PYTHON -m pip install . --no-deps -vv build_ext $BUILD_EXT_FLAGS install
```
fails with
```
pip._internal.exceptions.DistributionNotFound: No matching distribution found for build_ext
```

isuruf: You need something like,
```
$PYTHON -m pip install . --no-deps -vv --global-option=build_ext --global-option=--with-cuda 
```

pearu: Right. The following diff works locally:
```
diff --git a/recipe/build-pyarrow.sh b/recipe/build-pyarrow.sh
index 99396d8..4b01871 100644
--- a/recipe/build-pyarrow.sh
+++ b/recipe/build-pyarrow.sh
@@ -22,7 +22,7 @@ BUILD_EXT_FLAGS=""
 if [[ ! -z "${cuda_compiler_version+x}" && "${cuda_compiler_version}" != "None" ]]
 then
     export PYARROW_WITH_CUDA=1
-    BUILD_EXT_FLAGS="${BUILD_EXT_FLAGS} --with-cuda"
+    BUILD_EXT_FLAGS="${BUILD_EXT_FLAGS} --global-option=build_ext --global-option=--with-cuda"
 else
     export PYARROW_WITH_CUDA=0
 fi
@@ -35,7 +35,4 @@ fi
 
 cd python
 
-$PYTHON setup.py \
-        build_ext $BUILD_EXT_FLAGS \
-        install --single-version-externally-managed \
-                --record=record.txt
+$PYTHON -m pip install . --no-deps -vv $BUILD_EXT_FLAGS
diff --git a/recipe/meta.yaml b/recipe/meta.yaml
index 8a0edf7..20523e2 100644
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -181,6 +181,7 @@ outputs:
         - boost-cpp
         - cython
         - numpy 1.16.*
+        - pip
         - python
         - setuptools
         - setuptools_scm
```
Just FYI.

h-vetinari:
> @xhochy: Please do so at https://github.com/conda-forge/conda-forge.github.io

Followed-up in https://github.com/conda-forge/conda-forge.github.io/issues/1073.

PR #227, opened 2020-12-16
Fix osx-arm64 build

xhochy:
Checklist

### DIFF
--- a/recipe/7df87aca022ab6a309a6ee515370686aa3bcc013.patch
+++ b/recipe/7df87aca022ab6a309a6ee515370686aa3bcc013.patch
@@ -1,0 +1,32 @@
+From 7df87aca022ab6a309a6ee515370686aa3bcc013 Mon Sep 17 00:00:00 2001
+From: "Uwe L. Korn" <uwe.korn@quantco.com>
+Date: Mon, 14 Dec 2020 13:37:06 +0100
+Subject: [PATCH] ARROW-10873: [C++] Apple Silicon is reported as arm64 in
+ CMake
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+I'll add a CI entry soon that cross-compiles using `conda`, for now I keep submitting patches when things break for me.
+
+Closes #8893 from xhochy/ARROW-10873
+
+Authored-by: Uwe L. Korn <uwe.korn@quantco.com>
+Signed-off-by: Krisztián Szűcs <szucs.krisztian@gmail.com>
+---
+ cpp/cmake_modules/SetupCxxFlags.cmake | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/cpp/cmake_modules/SetupCxxFlags.cmake b/cpp/cmake_modules/SetupCxxFlags.cmake
+index 402d18f19d6..03d6a4ea34e 100644
+--- a/cpp/cmake_modules/SetupCxxFlags.cmake
++++ b/cpp/cmake_modules/SetupCxxFlags.cmake
+@@ -24,7 +24,7 @@ include(CheckCXXSourceCompiles)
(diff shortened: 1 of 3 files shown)


xhochy:
@jakirkham @kkraus14 @pearu @h-vetinari I unpinned CMake and now CUDA isn't found correctly anymore. Do you have suggestions on how to fix that?

```
-- Found Backtrace: /home/conda/feedstock_root/build_artifacts/arrow-cpp-ext_1608127253776/_build_env/x86_64-conda-linux-gnu/sysroot/usr/include  
CMake Error at /home/conda/feedstock_root/build_artifacts/arrow-cpp-ext_1608127253776/_build_env/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:218 (message):
  Could NOT find CUDA (missing: CUDA_INCLUDE_DIRS CUDA_CUDART_LIBRARY) (found
  version "9.2")
  /home/conda/feedstock_root/build_artifacts/arrow-cpp-ext_1608127253776/_build_env/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:577 (_FPHSA_FAILURE_MESSAGE)
  /home/conda/feedstock_root/build_artifacts/arrow-cpp-ext_1608127253776/_build_env/share/cmake-3.19/Modules/FindCUDA.cmake:1121 (find_package_handle_standard_args)
  src/arrow/gpu/CMakeLists.txt:32 (find_package)
```

jakirkham:
Is Arrow using [FindCUDA]( https://cmake.org/cmake/help/latest/module/FindCUDA.html )? If so, it may to use CUDA as a language support (analogous to C/C++) in CMake along with [`FindCUDAToolkit`]( https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#module:FindCUDAToolkit ) for specific libraries.

kkraus14:
> Is Arrow using [FindCUDA](https://cmake.org/cmake/help/latest/module/FindCUDA.html)? If so, it may to use CUDA as a language support (analogous to C/C++) in CMake along with [`FindCUDAToolkit`](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#module:FindCUDAToolkit) for specific libraries.

IIRC Arrow has no device code so it doesn't need to find `nvcc` in any way. It needs to just find the libs and headers for use in host code by the host compiler.

h-vetinari:
Had similar problems (IIRC) with cuda-detection in https://github.com/conda-forge/faiss-split-feedstock/pull/17 &  https://github.com/conda-forge/faiss-split-feedstock/pull/19.

In the the former, @beckermr, helped me with the following workaround (the surrounding comment contains a bunch more information):
https://github.com/conda-forge/faiss-split-feedstock/blob/master/recipe/build-lib.sh#L44-L52
Since then, there have been a couple of improvements to the nvcc-wrapper already (particularly on windows), but I tested only a few days ago to remove the workaround, and it still fails.

There's an additional wrinkle regarding `FindCUDA`, which @jaimergp mentioned in https://github.com/conda-forge/faiss-split-feedstock/pull/19:
> Unfortunately that verbose flag is for the deprecated `FindCUDA` module, not the current `FindCUDAToolkit`.

There, we ended up using `set "CudaToolkitDir=%CUDA_PATH%"`, but that was only relevant for Windows.

jaimergp:
AFAIK `openmm` also uses the deprecated `FindCUDA`, and we need (or needed) these two variables: https://github.com/conda-forge/openmm-feedstock/blob/master/recipe/build.sh#L19-L22

Maybe the stubs path is not needed anymore but I haven't checked. That should work (as of last week) although I haven't tried with last CMake yet.

jaimergp:
TIL about that [undocumented option](https://github.com/conda/conda-build/issues/2735) which was possibly behind https://github.com/conda-forge/nvcc-feedstock/issues/53 o.o

xhochy:
Seems like the culprit is hidden somewhere in `${CMAKE_ARGS}`. I'll investigate that in another PR.

PR #247, opened 2020-12-28
Windows + Cuda

h-vetinari:
Checklist

Not sure what's up with the bot having opened 3 PRs, but anyway, it's not smart enough for the specific skips here.

Closes #255

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set version = "2.0.0" %}
-{% set cuda_enabled = cuda_compiler_version is not undefined and cuda_compiler_version == '9.2' %}
+{% set cuda_enabled = cuda_compiler_version is not undefined and cuda_compiler_version != "None" %}
 {% set build_ext_version = "2.0.0" %}
 {% set build_ext = "cuda" if cuda_enabled else "cpu" %}
 {% set proc_build_number = "0" %}
@@ -17,8 +17,13 @@ source:
     - support-grpc-1.34.patch
 
 build:
-  number: 13
-  skip: true  # [cuda_compiler_version not in (undefined, "None", "9.2")]
+  number: 14
+  # for cuda on linux, building with 9.2 is enough to be compatible with all later versions,
+  # since arrow is only using libcuda, and not libcudart.
+  skip: true  # [linux and cuda_compiler_version not in ("None", "9.2")]
+  # same for cuda on windows (but starting from 10.0)
+  skip: true  # [win and cuda_compiler_version not in ("None", "10.0")]
+  skip: true  # [osx and cuda_compiler_version != "None"]
   run_exports:
     - {{ pin_subpackage("arrow-cpp", max_pin="x.x.x") }}
 
@@ -44,7 +49,6 @@ outputs:
     version: {{ version }}
     build:
       string: py{{ CONDA_PY }}h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}_{{ build_ext }}
-      skip: true  # [cuda_compiler_version not in (undefined, "None", "9.2")]
       run_exports:
         - {{ pin_subpackage("arrow-cpp", max_pin="x.x.x") }}
       ignore_run_exports:
@@ -161,7 +165,6 @@ outputs:
     version: {{ version }}
     build:
       string: py{{ CONDA_PY }}h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}_{{ build_ext }}
-      skip: true  # [cuda_compiler_version not in (undefined, "None", "9.2")]
       ignore_run_exports:
         - cudatoolkit
       track_features:
@@ -222,13 +225,13 @@ outputs:
       commands:
         - test ! -f ${SP_DIR}/pyarrow/tests/test_array.py       # [unix]
         - if exist %SP_DIR%/pyarrow/tests/test_array.py exit 1  # [win]
+
   - name: pyarrow-tests
     script: build-pyarrow.sh  # [not win]
     script: bld-pyarrow.bat   # [win]
     version: {{ version }}
     build:
       string: py{{ CONDA_PY }}h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}_{{ build_ext }}
-      skip: true  # [cuda_compiler_version not in (undefined, "None", "9.2")]
       ignore_run_exports:
         - cudatoolkit
       track_features:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -221,10 +221,15 @@ outputs:
         - pyarrow._hdfs
         # We can only test importing cuda package but cannot run when a
         # CUDA device is not available, for instance, when building from CI.
-        - pyarrow.cuda     # [cuda_compiler_version not in (undefined, "None")]
+        # On Windows, we cannot even do that due to `nvcuda.dll` not being found, see
+        # https://conda-forge.org/docs/maintainer/knowledge_base.html#nvcuda-dll-cannot-be-found-on-windows
+        # However, we check below for (at least) the presence of a correctly-compiled module
+        - pyarrow.cuda     # [cuda_compiler_version not in (undefined, "None") and not win]
       commands:
-        - test ! -f ${SP_DIR}/pyarrow/tests/test_array.py       # [unix]
-        - if exist %SP_DIR%/pyarrow/tests/test_array.py exit 1  # [win]
+        - test ! -f ${SP_DIR}/pyarrow/tests/test_array.py                         # [unix]
+        - if exist %SP_DIR%/pyarrow/tests/test_array.py exit 1                    # [win]
+        # Need to remove dot from PY_VER; %MYVAR:x=y% replaces "x" in %MYVAR% with "y"
+        - if not exist %SP_DIR%/pyarrow/_cuda.cp%PY_VER:.=%-win_amd64.pyd exit 1  # [win and cuda_compiler_version not in (undefined, "None")]
 
   - name: pyarrow-tests
     script: build-pyarrow.sh  # [not win]
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set version = "2.0.0" %}
-{% set cuda_enabled = cuda_compiler_version is not undefined and cuda_compiler_version != "None" %}
+{% set cuda_enabled = cuda_compiler_version != "None" %}
 {% set build_ext_version = "2.0.0" %}
 {% set build_ext = "cuda" if cuda_enabled else "cpu" %}
 {% set proc_build_number = "0" %}
@@ -70,7 +70,7 @@ outputs:
         - make  # [unix]
         - {{ compiler('c') }}
         - {{ compiler('cxx') }}
-        - {{ compiler("cuda") }}  # [cuda_compiler_version not in (undefined, "None")]
+        - {{ compiler("cuda") }}  # [cuda_compiler_version != "None"]
       host:
         - aws-sdk-cpp
         - boost-cpp >=1.70
@@ -99,7 +99,7 @@ outputs:
         - python
       run_constrained:
         - arrow-cpp-proc * {{ build_ext }}
-        - cudatoolkit >=9.2  # [cuda_compiler_version not in (undefined, "None")]
+        - cudatoolkit >=9.2  # [cuda_compiler_version != "None"]
 
     about:
       home: http://github.com/apache/arrow
@@ -181,7 +181,7 @@ outputs:
         - {{ compiler('c') }}
         - {{ compiler('cxx') }}
         # pyarrow does not require nvcc but it needs to link against libraries in arrow-cpp=*=*cuda
-        - {{ compiler("cuda") }}  # [cuda_compiler_version not in (undefined, "None")]
+        - {{ compiler("cuda") }}  # [cuda_compiler_version != "None"]
       host:
         - {{ pin_subpackage('arrow-cpp', exact=True) }}
         - cython
@@ -198,7 +198,7 @@ outputs:
         - python
       run_constrained:
         - arrow-cpp-proc * {{ build_ext }}
-        - cudatoolkit >=9.2  # [cuda_compiler_version not in (undefined, "None")]
+        - cudatoolkit >=9.2  # [cuda_compiler_version != "None"]
 
     about:
       home: http://github.com/apache/arrow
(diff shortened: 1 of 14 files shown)


xhochy:
You will also need to set `ARROW_CUDA=ON` in `bld-arrow.bat`.

h-vetinari:
Recording the cmake error here since azure will delete the logs
```
-- Found the Arrow core static library: ARROW_static_lib-NOTFOUND
-- Found ArrowPython: D:/bld/arrow-cpp-ext_1609201847987/_h_env/Library/include (found version "2.0.0") 
-- Found the Arrow Python by HOME: D:/bld/arrow-cpp-ext_1609201847987/_h_env/Library
-- Found the Arrow Python shared library: D:/bld/arrow-cpp-ext_1609201847987/_h_env/Library/bin/arrow_python.dll
-- Found the Arrow Python import library: D:/bld/arrow-cpp-ext_1609201847987/_h_env/Library/lib/arrow_python.lib
-- Found the Arrow Python static library: ARROW_PYTHON_static_lib-NOTFOUND
CMake Error at D:/bld/arrow-cpp-ext_1609201847987/_build_env/Library/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find ArrowCUDA (missing: ARROW_CUDA_INCLUDE_DIR
  ARROW_CUDA_LIB_DIR) (found version "2.0.0")
  D:/bld/arrow-cpp-ext_1609201847987/_build_env/Library/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:393 (_FPHSA_FAILURE_MESSAGE)
  cmake_modules/FindArrowCUDA.cmake:77 (find_package_handle_standard_args)
  CMakeLists.txt:405 (find_package)


-- Configuring incomplete, errors occurred!
See also "D:/bld/arrow-cpp-ext_1609201847987/work/python/build/temp.win-amd64-3.6/CMakeFiles/CMakeOutput.log".
See also "D:/bld/arrow-cpp-ext_1609201847987/work/python/build/temp.win-amd64-3.6/CMakeFiles/CMakeError.log".
```

Error looked familiar to problems we've had with https://github.com/conda-forge/faiss-split-feedstock/pull/19 (so I just tried bumping cmake), but it could also simply be that there's a problem on the arrow-side, particularly for correctly setting up `ARROW_CUDA_INCLUDE_DIR`, `ARROW_CUDA_LIB_DIR` on win.

recipe/meta.yaml
```diff
@@ -58,8 +62,8 @@ outputs:
         - numpy                                  # [build_platform != target_platform]
         - libtool                                # [osx and arm64]
         - libprotobuf
-        - grpc-cpp
```
xhochy, line 61: Any reason why this was deleted? This breaks the `osx-arm64` build.

xhochy:
The root issue is that `ARROW_CUDA` still is off in the C++ build.

```
--   ARROW_CUDA=OFF [default=OFF]
--       Build the Arrow CUDA extensions (requires CUDA toolkit)
```

recipe/meta.yaml
h-vetinari: Nope that was unintentional, sorry. Let's revert that commit (or rebase it out)

xhochy:
New failure:

```
  Syntax error in cmake code at

    D:/bld/arrow-cpp-ext_1609588880974/_build_env/Library/share/cmake-3.19/Modules/FindCUDA.cmake:760

  when parsing string

    CUDA_NVCC_EXECUTABLE;NAMES;nvcc;PATHS;C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.0;ENV;CUDA_PATH;ENV;CUDA_BIN_PATH;PATH_SUFFIXES;bin;bin64;NO_DEFAULT_PATH

  Invalid character escape '\P'.
  D:/bld/arrow-cpp-ext_1609588880974/_build_env/Library/share/cmake-3.19/Modules/FindCUDA.cmake:770 (cuda_find_host_program)
  src/arrow/gpu/CMakeLists.txt:32 (find_package)
```

xhochy:
@h-vetinari Can you have a look again? This now builds but fails the import tests. Probably related to the following output:

```
WARNING (arrow-cpp,Library/bin/arrow_cuda.dll): $RPATH/nvcuda.dll not found in packages, sysroot(s) nor the missing_dso_whitelist.
.. is this binary repackaging?
```

h-vetinari:
Hey @xhochy, I'm currently travelling without a laptop, should be able to take a look on the weekend.

h-vetinari:
> @h-vetinari Can you have a look again? This now builds but fails the import tests. Probably related to the following output:
> 
> ```
> WARNING (arrow-cpp,Library/bin/arrow_cuda.dll): $RPATH/nvcuda.dll not found in packages, sysroot(s) nor the missing_dso_whitelist.
> .. is this binary repackaging?
> ```

Found the following commit from @jaimergp regarding `nvcuda.dll`: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/commit/4bdbc451dc9546cf6e08264eaa085b5fe560e5b5, so that doesn't seem to be it (e.g. the openmm-feedstock has just [added](https://github.com/conda-forge/openmm-feedstock/commit/ebecada28c650f693e798fffde96fd43211648a2) it to the missing_dso_whitelist...).

Speaking of openmm, https://github.com/conda-forge/openmm-feedstock/commit/ebecada28c650f693e798fffde96fd43211648a2 might be interesting, in that it copies `nvcuda.dll` (from where?) to system32.

jaimergp:
I was trying to copy the pre-dll from the CUDA installer directly, but it didn't work. I guess there's a post-processing happening somewhere that converts the pre-dll into a valid dll. Anyway, since there's no GPU in the CI machine, the drivers won't be fully installed. Check this comment: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/master/recipe/install_cuda.bat#L140

Probably worth adding to the documentation, now that I think about it. Thanks!

xhochy:
Reading https://github.com/conda-forge/conda-forge.github.io/pull/1226 tells me that we probably should skip the import test for `pyarrow._cuda` on Windows, is this correct? 

Persisting the error log here:

```
import: 'pyarrow.cuda'
  File "D:\bld\arrow-cpp-ext_1610055721817\test_tmp\run_test.py", line 26, in <module>
    import pyarrow.cuda
  File "D:\bld\arrow-cpp-ext_1610055721817\_test_env\lib\site-packages\pyarrow\cuda.py", line 21, in <module>
    from pyarrow._cuda import (Context, IpcMemHandle, CudaBuffer,
ImportError: DLL load failed: The specified module could not be found.
```

h-vetinari:
> @xhochy: Reading [conda-forge/conda-forge.github.io#1226](https://github.com/conda-forge/conda-forge.github.io/pull/1226) tells me that we probably should skip the import test for `pyarrow._cuda` on Windows, is this correct?

If the import of `pyarrow._cuda` already requires `nvcuda.dll`, then I'd say so.

xhochy:
> > @xhochy: Reading [conda-forge/conda-forge.github.io#1226](https://github.com/conda-forge/conda-forge.github.io/pull/1226) tells me that we probably should skip the import test for `pyarrow._cuda` on Windows, is this correct?
> 
> If the import of `pyarrow._cuda` already requires `nvcuda.dll`, then I'd say so.

Cool, then I add this and rebase.

h-vetinari:
> Cool, then I add this and rebase.

Sorry, was already doing this before I saw your response

xhochy:
> > Cool, then I add this and rebase.
> 
> Sorry, was already doing this before I saw your response

I can wait ;)

xhochy:
@h-vetinari Do you want to finish this (I though you were still without a laptop :D)?

h-vetinari:
> @h-vetinari Do you want to finish this (I thought you were still without a laptop :D)?

Weekend came early. ;-)

After your "I can wait", I had a look at adding `and not win` in the selector for the test, but the comment there made me pause for a second. Not sure if something happened on the arrow side of things, but the comment (from @pearu in #146) implies that `import` _should_ actually work (while actually running code will fail due to lack of hardware):
```python
# We can only test importing cuda package but cannot run when a
# CUDA device is not available, for instance, when building from CI.
```

h-vetinari:
Alright, I think I untwisted my thoughts on this successfully - added a comment in https://github.com/conda-forge/arrow-cpp-feedstock/pull/247/commits/7647fde52b22b34324078f0f23a8dd63b4fe82be; PTAL @xhochy @jaimergp @pearu

h-vetinari:
@xhochy, please don't merge this yet, even if the CI should turn green.

I want to try removing a bunch of the debug stuff after conda-forge/nvcc-feedstock#58 (and also squash a bit - e.g. https://github.com/conda-forge/arrow-cpp-feedstock/pull/247/commits/5b573af1b545af5136ac69970e35a95d599fcf81 has a wrong commit message)

recipe/meta.yaml
```diff
@@ -218,17 +221,19 @@ outputs:
         - pyarrow._hdfs
         # We can only test importing cuda package but cannot run when a
         # CUDA device is not available, for instance, when building from CI.
-        - pyarrow.cuda     # [cuda_compiler_version not in (undefined, "None")]
+        # On Windows, we cannot even test the import due to `nvcuda.dll` not
+        # being found, see conda-forge/conda-forge.github.io#1226
+        - pyarrow.cuda     # [cuda_compiler_version not in (undefined, "None") and not win]
```
xhochy, line 226: Can we check instead for the existence of the `pyarrow._cuda` Python module then? So we can at least ensure that it was built.

h-vetinari: What's the difference between `pyarrow._cuda` and `pyarrow.cuda`? If the former can be imported without needing `nvcuda.dll`, that would be an easy solution (and we could still test importing `pyarrow.cuda` on linux).

xhochy: `pyarrow.cuda` is a bit of Python boilerplate and `pyarrow._cuda` is the compiled Cython code linking to `libarrow_cuda`. Both need `nvcuda.dll` to work as `pyarrow.cuda` depends on `pyarrow._cuda`.

h-vetinari: OK, there's not much "glue" between `_cuda` and `cuda`: https://github.com/apache/arrow/blob/master/python/pyarrow/cuda.py

> @xhochy: Can we check instead for the existence of the `pyarrow._cuda` Python module then? So we can at least ensure that it was built.

I have to say I'm unable to identify which part of https://github.com/apache/arrow/blob/master/python/pyarrow/_cuda.pyx would trigger the DLL load (or is that added during compilation? shouldn't that by cython-only?) - but maybe it's possible to import a subset of:
```
from pyarrow._cuda import (Context, IpcMemHandle, CudaBuffer,
                           HostBuffer, BufferReader, BufferWriter,
                           new_host_buffer,
                           serialize_record_batch, read_message,
                           read_record_batch)
```

h-vetinari: Sorry, GH didn't show me your reply so I went searching myself. The part where I don't understand from the source of `_cuda.pyx` where the DLL-dependence comes in is still relevant though.

h-vetinari:
@jaimergp
Unfortunately, the [CI](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=260135&view=results) for https://github.com/conda-forge/arrow-cpp-feedstock/pull/247/commits/fdbbc0cb99b750ac8cfbb488246e67b0e0c56c53 shows that conda-forge/nvcc-feedstock#58 does not yet allow removing the cuda-debug stuff in this recipe.

In particular, it seems it's necessary to switch all backslashes to forward slashes, as shown here:
```
Syntax error in cmake code at
    D:/bld/arrow-cpp-ext_1610120989927/_build_env/Library/share/cmake-3.19/Modules/FindCUDA.cmake:760
  when parsing string
    CUDA_NVCC_EXECUTABLE;NAMES;nvcc;PATHS;C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.0;ENV;CUDA_PATH;ENV;CUDA_BIN_PATH;PATH_SUFFIXES;bin;bin64;NO_DEFAULT_PATH
  Invalid character escape '\P'.
```

As soon as the CI is through I'll push another commit to test this hypothesis.

jaimergp:
Thanks for the report! I will add another PR for `nvcc` then! Slowly getting there... Thanks for the patience :)

jaimergp:
Does it go away if you do something like:

```
set CMAKE_ARGS=%CMAKE_ARGS:\=/%
```

?

h-vetinari:
> Thanks for the report! I will add another PR for `nvcc` then! Slowly getting there... Thanks for the patience :)

Cool, great to hear, thanks! :)

> Does it go away if you do something like:
> 
> ```
> set CMAKE_ARGS=%CMAKE_ARGS:\=/%
> ```
> 
> ?

The only "debug" stuff that's still necessary is:
```batch
    REM this should move to nvcc-feedstock
    set "CUDA_PATH=%CUDA_PATH:\=/%"
    set "CUDA_HOME=%CUDA_HOME:\=/%"
```

jaimergp:
Oh, sorry, I got confused by the `-DCUDA_TOOLKIT_ROOT_DIR=%CUDA_PATH:\=/%"` part.

So this suggests that we should export `CUDA_PATH` with forward slashes only, I see. Can you check if CUDA_HOME is necessary? I think CMake only relies on CUDA_PATH.

h-vetinari:
> So this suggests that we should export `CUDA_PATH` with forward slashes only, I see. Can you check if CUDA_HOME is necessary? I think CMake only relies on CUDA_PATH.

In the history I just rebased, @xhochy added this for a reason, but I'll try to make sure, after the CI has run through

xhochy:
> > So this suggests that we should export `CUDA_PATH` with forward slashes only, I see. Can you check if CUDA_HOME is necessary? I think CMake only relies on CUDA_PATH.
> 
> In the history I just rebased, @xhochy added this for a reason, but I'll try (just to make sure) after the CI has run through

We need `CUDA_HOME` because of this Arrow-specific behaviour: https://github.com/apache/arrow/blob/a576cd71c08c3fec029d24092c407ece1775f5a6/cpp/src/arrow/gpu/CMakeLists.txt#L28-L30

jaimergp:
I see, thank you. I will link the relevant PR here once is ready for comments.

recipe/meta.yaml
```diff
@@ -218,17 +221,19 @@ outputs:
...
       commands:
         - test ! -f ${SP_DIR}/pyarrow/tests/test_array.py       # [unix]
         - if exist %SP_DIR%/pyarrow/tests/test_array.py exit 1  # [win]
```
xhochy, line 229:
```suggestion
        - if exist %SP_DIR%/pyarrow/tests/test_array.py exit 1  # [win]
        - if not exist %SP_DIR%/pyarrow/_cuda.cp%PY_VER%-win_amd64.pyd exit 1  # [cuda_compiler_version not in (undefined, "None") and win]
```

h-vetinari: Incorporated into https://github.com/conda-forge/arrow-cpp-feedstock/pull/247/commits/43d3175bde1ddcb1468821b1191f275a032f82a0, thanks!

h-vetinari:
@xhochy, I think this is ready :)

xhochy (approved):

xhochy:
🎉

PR #387, opened 2021-03-23
Pinning libprotobuf=3.6.*

gabrielcnr:
Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -16,7 +16,7 @@ source:
     - 9569.patch
 
 build:
-  number: 9
+  number: 10
   # for cuda on win/linux, building with 9.2 is enough to be compatible with all later versions,
   # since arrow is only using libcuda, and not libcudart.
   skip: true  # [(win or linux) and cuda_compiler_version not in ("None", "10.2")]
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -59,7 +59,7 @@ outputs:
         - cython                                 # [build_platform != target_platform]
         - numpy                                  # [build_platform != target_platform]
         - gnuconfig                              # [osx and arm64]
-        - libprotobuf
+        - libprotobuf =3.6.*
         - grpc-cpp
         - cmake
         - autoconf  # [unix]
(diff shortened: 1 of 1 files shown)


gabrielcnr:
@marcelotrevisani - as discussed, this fixes an `ImportError` when picking up `>=3.6`

xhochy (changes requested):
No, we're not diverting from the global pinning here.

xhochy:
@gabrielcnr Can you please provide further context here?

gabrielcnr:
Had this problem in a complex environment I have here... subset of the problem can be reproduced by creating an environment like the one below:

```
mamba create -n ibis ibis-framework==0.14.0 python=3.6 “pyarrow>=0.13.0,<0.14” sqlalchemy=1.3 -y -c Conda-forge
```

And then trying to `import ibis` in Python will crash with an ImportError because of protobuf...

xhochy:
That is a really old `pyarrow` version, that will not be fixed by breaking the 3.0 build.

xhochy:
@gabrielcnr This rather seems like one of the packages you have installed which is not `pyarrow` has a dependency on `protobuf` *and* `libprotobuf` but did only declare on the first and thus didn't get a strict pin.

gabrielcnr:
@xhochy - it makes sense, it's an old `pyarrow` and won't be able to upgrade quickly right now. I'm happy to pin `libprotobuf` to be `3.6.*` on the target environment. This solves my problem. The PR can be closed then. Thanks.

xhochy:
I'm a bit confused though as `pyarrow` should already have such a strict `libprotobuf` pin.

marcelotrevisani:
I believe we can merge that in a separated branch for that pyarrow version, it will fix it

xhochy:
We have a pin for `https://github.com/conda-forge/arrow-cpp-feedstock/blob/d8bc9d0450b2e053fddf9eab481614a0da37609d/.ci_support/linux_python3.6.yaml#L17-L18` `libprotobuf=3.7`. So I'm confused why it works with an older version.

marcelotrevisani:
that is odd indeed. I thought the older version was using protobuf 3.6

marcelotrevisani:
unnn... maybe it is getting that package that was generated from the archived feedstock?

marcelotrevisani:
> unnn... maybe it is getting that package that was generated from the archived feedstock?

That is not the case. I just verified. Yeah... very intriguing

PR #418, opened 2021-05-08
Remove aws-sdk-cpp runtime dependency

wietsedv:
`aws-sdk-cpp` is a runtime dependency, but it should not be required to use Apache Arrow and `pyarrow` specifically. I'm not sure whether I can simply remove it without any consequences. Maybe it should just be a build dependency.


xhochy:
What makes you think it isn't one? The S3 support links to it.

xhochy:
PyPi packages don't have native dependency management. They all come bundled with the wheel. This is the major difference between simply using pip and using conda: The solver and the build system take also care of your non-Python dependencies.

PR #468, opened 2021-06-22
Readd `1.0.x` migrations

jakirkham:
Currently in RAPIDS we are still on Arrow 1.0. Have run into segfaults when trying to upgrade to 4.0 ( https://github.com/rapidsai/cudf/pull/7495 ). Our most recent release is still on Arrow 1.0 for this reason. However there are some conflicts when installing (particularly around libthrift version 0.14.2). Try to resuscitate this branch to clear up those issues for users trying to install the last release.

<hr>

Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

PR #469, opened 2021-06-22
Refresh 1.0.x

jakirkham:
Re-rendered with conda-build 3.21.4, conda-smithy 3.10.3, and conda-forge-pinning 2021.06.22.04.56.18 in response to this suggestion ( https://github.com/conda-forge/arrow-cpp-feedstock/pull/468#issuecomment-865548085 ) and applied all recent migrations

Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)


jakirkham:
@conda-forge-admin, please re-render

jakirkham:
It looks like the tarballs just got removed from the distribution server. @xhochy do you know what is going on here?

jakirkham:
Switched to a GH archive to workaround the issue above

jakirkham:
It appears Conda runs into conflicts when trying to install `lz4-c` version `1.9.2` on Windows. However relaxing that constraint results in an inability to find the LZ4 library

recipe/bld-arrow.bat
```diff
@@ -46,8 +46,18 @@ cmake -G "Ninja" ^
...
+) else (
+	goto :cfgfail
+)
+
+:cfgfail
+type CMakeFiles\CMakeOutput.log
+type CMakeFiles\CMakeError.log
+exit 1
 
+:build
```
jakirkham, line 60: There's probably a better way to accomplish this, but I lack a Windows machine to test this on. Would welcome input from others :)

jakirkham:
All the other builds pass, but I probably can't figure out the Windows issues myself (I lack a machine to do so)

xhochy:
The changes from https://github.com/apache/arrow/commit/81d3f2657b17436d6d5a6af9aaf6f36c3f5e4ac9#diff-4dc66d03c35d7df04797077cdb9794df6825428d7fa06438cefe037025e83dd6 might help here. Maybe setting `-DLZ4_MSVC_STATIC_LIB_SUFFIX=""` is already sufficient.

jakirkham:
Thanks for the suggestion Uwe! 😄 Seems to be working 🍀  🎉

jakirkham:
Looks like we are running into a new issue with Gandiva. It finds the dynamic library, but it seems to error because the static library is missing. Not sure if there's an easy way to disable that

xhochy:
The actual issue is:

```
-- Detecting CXX compile features - done
CMake Error at CMakeLists.txt:95 (include):
  include could not find requested file:

    BuildUtils


CMake Error at CMakeLists.txt:100 (include):
  include could not find requested file:

    SetupCxxFlags
```

recipe/meta.yaml
```diff
@@ -9,8 +9,8 @@ package:
   version: {{ version }}
 
 source:
-  url: https://dist.apache.org/repos/dist/release/arrow/arrow-{{ version }}/apache-arrow-{{ version }}.tar.gz
-  sha256: 149ca6aa969ac5742f3b30d1f69a6931a533fd1db8b96712e60bf386a26dc75c
+  url: https://github.com/apache/arrow/archive/refs/tags/apache-arrow-{{ version }}.tar.gz
```
xhochy, line 12: This one should work, I cannot commit to this PR though
```suggestion
  url: http://archive.apache.org/dist/arrow/arrow-{{ version }}/apache-arrow-{{ version }}.tar.gz
```

xhochy: This one might also fix the build!

jakirkham: Indeed it has. Thanks for pointing this out 😄

So are there additional files that are generated in that archive? Would we need to generate them some how ourselves when using the git repo as a source? (mostly asking for my own understanding)

xhochy: They are copied in the release archive. In the git repo they are symlinked but the symlink between the cpp and python directory is not reflected in the Github archive.

jakirkham: Ah ok. So if we cloned the repo instead, it would have worked?

PR #839, opened 2022-09-08
Join as maintainer to take care of migrations

h-vetinari:
Hey all

There's currently a few tricky migrations pending (abseil, grpc, libprotobuf), and since arrow plays such a central role in the ecosystem nowadays, it's good to keep this moving along across all ABI branches. Not least since I also help co-maintain those other feedstocks, I'd be happy to help take care of these migrations, and this would be easier as a maintainer (can push into bot PRs, no need for attention on trivial PRs).

I thought I'd propose joining first (considering the quantity of open migration PRs), but if desired I can also try raising independent PRs first.

PS. I'm not completely new to this feedstock, c.f. https://github.com/conda-forge/arrow-cpp-feedstock/pull/247 & collab on https://github.com/conda-forge/pyarrow-feedstock/pull/101. :)

PPS. A standard rerender seems to bring in the (now-closed) orc migration; if desired I can undo this part.

Closes #826

PR #840, opened 2022-09-08
[main] Rebuild for abseil_cpp202206230

h-vetinari:
Builds on top of #839, will be rebased once that's in.


h-vetinari:
This is green (aside from regular timeouts). I'm going to push to main directly to avoid another huge CI run, as I'm planning to economize those, but want to also start working on the other branches.

The last commit here is 614fa1ea2b4aa86a344c876913f22076fd88b163, the commit I'm pushing is c507b095e4d39c8430da1c0e988bf49f49a13135, and one can verify that the diff between them is empty.

Issue #567, opened 2021-11-08
aws error after writing parquet file


kylekeppler:
- [x] I read [the conda-forge documentation](https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge) and could not find the solution for my problem there.

Issue:

After converting a dataframe to parquet, I get an aws-related error:

```
Fatal error condition occurred in D:\bld\aws-c-io_1633633258269\work\source\event_loop.c:74: aws_thread_launch(&cleanup_thread, s_event_loop_destroy_async_thread_fn, el_group, &thread_options) == AWS_OP_SUCCESS
Exiting Application
at 0x7FFFB9D94380: aws_backtrace_print
at 0x7FFFB9D83560: aws_fatal_assert
at 0x7FFFA8FB5F10: aws_event_loop_wait_for_stop_completion
at 0x7FFFB9D91470: aws_ref_count_release
at 0x7FFFA8FB3D80: aws_server_bootstrap_set_alpn_callback
at 0x7FFFB9D91470: aws_ref_count_release
at 0x7FFFA8FB3760: aws_client_bootstrap_release
at 0x7FFFC31976F0: Aws::Crt::Io::ClientBootstrap::~ClientBootstrap
at 0x7FFFC2F7EB40: Aws::Utils::Stream::SimpleStreamBuf::xsputn
at 0x7FFFD9E71BA0: execute_onexit_table
at 0x7FFFD9E71BA0: execute_onexit_table
at 0x7FFFD9E71BA0: execute_onexit_table
at 0x7FFFC2F7EB40: Aws::Utils::Stream::SimpleStreamBuf::xsputn
at 0x7FFFC2F7EB40: Aws::Utils::Stream::SimpleStreamBuf::xsputn
at 0x7FFFDCD55AA0: RtlAnsiStringToUnicodeString
at 0x7FFFDCD590F0: LdrShutdownProcess
at 0x7FFFDCD9CAB0: RtlExitUserProcess
at 0x7FFFDB4ED620: ExitProcess
at 0x7FFFD9E80C50: exit
at 0x7FFFD9E80C50: exit
at 0x7FF60DDA1160: OPENSSL_Applink
at 0x7FFFDB4E7960: BaseThreadInitThunk
at 0x7FFFDCD8A2D0: RtlUserThreadStart
```

[...]

PR #880, opened 2022-10-27
WIP: try unpinning aws-sdk-cpp

h-vetinari:
See #567 / https://github.com/aws/aws-sdk-cpp/issues/1809 / https://issues.apache.org/jira/browse/ARROW-15141

Fixing this is necessary to unblock openssl 3 migration, because 1.8.x is not compatible with the newer aws-c-* stack anymore, see https://github.com/conda-forge/aws-sdk-cpp-feedstock/pull/577


h-vetinari:
So I was just wondering - how could this supposed reproducer be exercising any aws-related code? Presumably only if it's executed _on_ AWS? 🤔

Issue #695, opened 2022-03-30
arrow-cpp 7.0.0 is incompatible with tensorflow-gpu (fails with requires libprotobuf >=3.19.4,<3.20.0a0)


holzman:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

arrow-cpp 7.0.0 is not compatible with tensorflow-gpu (arrow-cpp 6.0.1 works fine).

To reproduce:
```
FROM centos:centos7
ARG MAMBAFORGE=https://github.com/conda-forge/miniforge/releases/download/4.11.0-4/Mambaforge-Linux-x86_64.sh
RUN curl -sSL $MAMBAFORGE -o /tmp/mambaforge.sh && bash /tmp/mambaforge.sh -bfp /usr/local
RUN mamba install --yes arrow-cpp==7.0.0 tensorflow-gpu
```

The error:
```
Encountered problems while solving:
  - package arrow-cpp-7.0.0-py39h116028d_0_cpu requires libprotobuf >=3.19.4,<3.20.0a0, but none of the providers can be installed
```

I have another configuration (that I haven't simplified, so reproducer not attached) that complains similarly:
```
Encountered problems while solving:
  - package arrow-cpp-7.0.0-py37h1be195e_2_cuda requires libprotobuf >=3.19.4,<3.20.0a0, but none of the providers can be installed
```

### Installed packages

```shell
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       1_gnu    conda-forge
brotlipy                  0.7.0           py39h3811e60_1003    conda-forge
bzip2                     1.0.8                h7f98852_4    conda-forge
c-ares                    1.18.1               h7f98852_0    conda-forge
ca-certificates           2021.10.8            ha878542_0    conda-forge
certifi                   2021.10.8        py39hf3d152e_1    conda-forge
cffi                      1.15.0           py39h4bc2ebd_0    conda-forge
charset-normalizer        2.0.12             pyhd8ed1ab_0    conda-forge
colorama                  0.4.4              pyh9f0ad1d_0    conda-forge
conda                     4.11.0           py39hf3d152e_0    conda-forge
conda-package-handling    1.7.3            py39h3811e60_1    conda-forge
cryptography              36.0.1           py39h95dcef6_0    conda-forge
icu                       69.1                 h9c3ff4c_0    conda-forge
[...]

PR #896, opened 2022-12-04
[RFC] Keep building for py37 

h-vetinari:
This was explicitly [requested](https://github.com/conda-forge/arrow-cpp-feedstock/pull/875#issuecomment-1332753403) by @rxm7706, and also [asked about](https://github.com/conda-forge/arrow-cpp-feedstock/pull/875#pullrequestreview-1199167120) by upstream dev @raulcd in the same PR.

I said at the time:
> It's technically possible. Assuming the CI passes it would not add an extreme maintenance burden, though it depends also whether the other maintainers agree (w.r.t. precedent etc.). But this will be a topic for a separate PR.

This PR is to see if it passes CI cleanly (thankfully, the host deps don't carry much python-deps that are migration-sensitive, so this might just work), and then gather feedback from @conda-forge/arrow-cpp if we actually want to do this.

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -1,0 +1,17 @@
+# exceptionally reinstate python 3.7 support, to be removed
+# at any time, but ideally once upstream drops support itself
+python:
+  - 3.7.* *_cpython   # [not (osx and arm64)]
+  - 3.8.* *_cpython
+  - 3.9.* *_cpython
+  - 3.10.* *_cpython
+python_impl:
+  - cpython           # [not (osx and arm64)]
+  - cpython
+  - cpython
+  - cpython
+numpy:
+  - 1.20              # [not (osx and arm64)]
+  - 1.20
+  - 1.20
+  - 1.21
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -5,8 +5,10 @@ python:
   - 3.8.* *_cpython
   - 3.9.* *_cpython
   - 3.10.* *_cpython
+  - 3.11.* *_cpython
 python_impl:
   - cpython           # [not (osx and arm64)]
+  - cpython
   - cpython
   - cpython
   - cpython
@@ -15,3 +17,4 @@ numpy:
   - 1.20
   - 1.20
   - 1.21
+  - 1.23
--- a/.ci_support/linux_64_cuda_compiler_version10.2openssl1.1.1.yaml
+++ b/.ci_support/linux_64_cuda_compiler_version10.2openssl1.1.1.yaml
@@ -45,6 +45,7 @@ numpy:
 - '1.23'
 - '1.20'
 - '1.20'
+- '1.20'
 openssl:
 - 1.1.1
 orc:
@@ -56,6 +57,7 @@ pin_run_as_build:
 python:
 - 3.10.* *_cpython
 - 3.11.* *_cpython
+- 3.7.* *_cpython
 - 3.8.* *_cpython
 - 3.9.* *_cpython
 re2:
--- a/.ci_support/linux_64_cuda_compiler_version10.2openssl3.yaml
+++ b/.ci_support/linux_64_cuda_compiler_version10.2openssl3.yaml
@@ -45,6 +45,7 @@ numpy:
 - '1.23'
 - '1.20'
 - '1.20'
+- '1.20'
 openssl:
 - '3'
 orc:
@@ -56,6 +57,7 @@ pin_run_as_build:
 python:
 - 3.10.* *_cpython
 - 3.11.* *_cpython
+- 3.7.* *_cpython
 - 3.8.* *_cpython
 - 3.9.* *_cpython
 re2:
--- a/.ci_support/linux_64_cuda_compiler_versionNoneopenssl1.1.1.yaml
+++ b/.ci_support/linux_64_cuda_compiler_versionNoneopenssl1.1.1.yaml
@@ -45,6 +45,7 @@ numpy:
 - '1.23'
 - '1.20'
 - '1.20'
+- '1.20'
 openssl:
 - 1.1.1
 orc:
@@ -56,6 +57,7 @@ pin_run_as_build:
 python:
 - 3.10.* *_cpython
 - 3.11.* *_cpython
+- 3.7.* *_cpython
 - 3.8.* *_cpython
 - 3.9.* *_cpython
 re2:
--- a/.ci_support/linux_64_cuda_compiler_versionNoneopenssl3.yaml
+++ b/.ci_support/linux_64_cuda_compiler_versionNoneopenssl3.yaml
@@ -45,6 +45,7 @@ numpy:
 - '1.23'
 - '1.20'
 - '1.20'
+- '1.20'
 openssl:
 - '3'
 orc:
@@ -56,6 +57,7 @@ pin_run_as_build:
 python:
 - 3.10.* *_cpython
 - 3.11.* *_cpython
+- 3.7.* *_cpython
 - 3.8.* *_cpython
 - 3.9.* *_cpython
 re2:
--- a/.ci_support/linux_aarch64_cuda_compiler_versionNoneopenssl1.1.1.yaml
+++ b/.ci_support/linux_aarch64_cuda_compiler_versionNoneopenssl1.1.1.yaml
@@ -45,6 +45,7 @@ numpy:
 - '1.23'
 - '1.20'
 - '1.20'
+- '1.20'
 openssl:
 - 1.1.1
 orc:
@@ -56,6 +57,7 @@ pin_run_as_build:
 python:
 - 3.10.* *_cpython
 - 3.11.* *_cpython
+- 3.7.* *_cpython
 - 3.8.* *_cpython
 - 3.9.* *_cpython
 re2:
--- a/.ci_support/linux_aarch64_cuda_compiler_versionNoneopenssl3.yaml
+++ b/.ci_support/linux_aarch64_cuda_compiler_versionNoneopenssl3.yaml
@@ -45,6 +45,7 @@ numpy:
 - '1.23'
 - '1.20'
 - '1.20'
+- '1.20'
 openssl:
 - '3'
 orc:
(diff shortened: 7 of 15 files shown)


h-vetinari:
So the problem here is that despite https://github.com/conda-forge/arrow-cpp-feedstock/pull/896/commits/3d7347341f88c55021541e43289b5684dc9add53 being purely additive (in terms of config), it somehow disables the builds for py311:
```
anaconda upload \
    /home/conda/feedstock_root/build_artifacts/linux-64/apache-arrow-proc-4.0.0-cuda.conda \
    /home/conda/feedstock_root/build_artifacts/linux-64/libarrow-10.0.1-h9ff0032_0_cuda.conda \
    /home/conda/feedstock_root/build_artifacts/linux-64/arrow-cpp-10.0.1-h3e2b116_0_cuda.conda \
    /home/conda/feedstock_root/build_artifacts/linux-64/arrow-cpp-proc-4.0.0-cuda.conda \
    /home/conda/feedstock_root/build_artifacts/linux-64/pyarrow-10.0.1-py37he8182db_0_cuda.conda \
    /home/conda/feedstock_root/build_artifacts/linux-64/pyarrow-10.0.1-py310ha7322a8_0_cuda.conda \
    /home/conda/feedstock_root/build_artifacts/linux-64/pyarrow-10.0.1-py39h4d4a1a0_0_cuda.conda \
    /home/conda/feedstock_root/build_artifacts/linux-64/pyarrow-10.0.1-py38h1fd4ed9_0_cuda.conda \
    /home/conda/feedstock_root/build_artifacts/linux-64/pyarrow-tests-10.0.1-py37h82fcdeb_0_cuda.conda \
    /home/conda/feedstock_root/build_artifacts/linux-64/pyarrow-tests-10.0.1-py310h50e40b8_0_cuda.conda \
    /home/conda/feedstock_root/build_artifacts/linux-64/pyarrow-tests-10.0.1-py39h947458c_0_cuda.conda \
    /home/conda/feedstock_root/build_artifacts/linux-64/pyarrow-tests-10.0.1-py38hbf57a63_0_cuda.conda
anaconda_upload is not set.  Not uploading wheels: []
```

This has nothing to do with this feedstock, but is a problem of `conda-build` it seems. FWIW, I've tried the following patch to cbc, and nothing changes (the rerender is already correct):
```diff
diff --git a/recipe/conda_build_config.yaml b/recipe/conda_build_config.yaml
index cd4b5d8..ed8457f 100644
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -5,13 +5,16 @@ python:
   - 3.8.* *_cpython
   - 3.9.* *_cpython
   - 3.10.* *_cpython
+  - 3.11.* *_cpython
 python_impl:
   - cpython           # [not (osx and arm64)]
   - cpython
   - cpython
   - cpython
+  - cpython
 numpy:
   - 1.20              # [not (osx and arm64)]
   - 1.20
   - 1.20
   - 1.21
+  - 1.23
```

As it stands, someone can fix what's causing the missing py311 builds (not me), or this PR is dead.

rxm7706:
> So the problem here is that despite [3d73473](https://github.com/conda-forge/arrow-cpp-feedstock/commit/3d7347341f88c55021541e43289b5684dc9add53) being purely additive (in terms of config), it somehow disables the builds for py311:

@conda-forge/help-python could you help here, is this a knows issue with a workaround.

recipe/conda_build_config.yaml
```diff
@@ -0,0 +1,17 @@
...
+python_impl:
+  - cpython           # [not (osx and arm64)]
+  - cpython
+  - cpython
+  - cpython
+numpy:
+  - 1.20              # [not (osx and arm64)]
+  - 1.20
+  - 1.20
+  - 1.21
```
mbargull, line 17:
```suggestion
# exceptionally reinstate python 3.7 support, to be removed
# at any time, but ideally once upstream drops support itself
python:
  - 3.7.* *_cpython   # [not (osx and arm64)]
  - 3.8.* *_cpython
  - 3.9.* *_cpython
  - 3.10.* *_cpython
  - 3.11.* *_cpython
python_impl:
  - cpython           # [not (osx and arm64)]
  - cpython
  - cpython
  - cpython
  - cpython
numpy:
  - 1.20              # [not (osx and arm64)]
  - 1.20
  - 1.20
  - 1.21
  - 1.23
```

mbargull:

h-vetinari:
Hey @mbargull, thanks for taking a look!

> FWIW, I've tried the following patch to cbc \[equivalent to https://github.com/conda-forge/arrow-cpp-feedstock/pull/896/commits/55678c77c8f0cb8f1f63e9ebdded571e7caf45e5 ], and nothing changes (the rerender is already correct):

As I mentioned, it's not a question of the rerendering; https://github.com/conda-forge/arrow-cpp-feedstock/pull/896/commits/3d7347341f88c55021541e43289b5684dc9add53 already has the correct set of python versions to build.

mbargull:
> As I mentioned, it's not a question of the rerendering; [3d73473](https://github.com/conda-forge/arrow-cpp-feedstock/commit/3d7347341f88c55021541e43289b5684dc9add53) already has the correct set of python versions to build.

Yes, I don't expect the rerendering to yield any changes. I only set off that bot request for good measure.
For the actual explanation why 3.11 variants weren't build and why you need 3.1 in cbc.yaml, see https://github.com/conda/conda-build/issues/4666#issuecomment-1336398776 :
> Values from the recipe's `conda_build_config.yaml` take precedence over those from other config files.
Note they are not additive but replace higher priority ones.

mbargull:
> I said at the time:
> 
> > It's technically possible. Assuming the CI passes it would not add an extreme maintenance burden, though it depends also whether the other maintainers agree (w.r.t. precedent etc.). But this will be a topic for a separate PR.

Yes, this can indeed create additional maintenance burden in case dependencies are updated/rebuilt.
Since we do not build for Python 3.7 by default anymore on conda-forge any Python package dependencies will likely be stuck on older versions if their newer versions are built without Python 3.7 variants.
(This feedstock doesn't seem to have too many Python package dependencies, so probably not a big deal; but just a thing to generally keep in mind.)

h-vetinari:
> For the actual explanation why 3.11 variants weren't build and why you need 3.11 in cbc.yaml, see [conda/conda-build#4666 (comment)](https://github.com/conda/conda-build/issues/4666#issuecomment-1336398776) :

Ah, so I got tricked into ignoring that diff because the rerender didn't change, but conda build still needs it, got it. Thanks for looking at this so quickly!

> Since we do not build for Python 3.7 by default anymore on conda-forge any Python package dependencies will likely be stuck on older versions if their newer versions are built without Python 3.7 variants.

I'm aware; I also looked at the deps and they don't look too abi-migration-exposed, but yeah, my standpoint is essentially: python 3.7 is dead in conda-forge, but as long as it only adds a small amount of CI runtime and no other complications, I'm fine with building those packages in the fact of an explicit request (also, since we're now running the test suite, it's unlikely that broken packages wouldn't sneak past).

h-vetinari:
@conda-forge/arrow-cpp 
There's an explicit user request to provide python 3.7 builds for a while longer. With the new CI setup it doesn't cost much (just a few minutes runtime within the same job), so I'm leaning towards accommodating that request.

If there are any opposing voices, please speak up, otherwise I'm planning to merge this in a week. :)

xhochy:
Fine with me, we probably want to drop this in 6-12 months latest though ;)

jakirkham:
I'd prefer skipping this. Conda-forge already dropped it. We will eventually start feeling the pain of not having other Python 3.7 packages not get updated. Easier to let it go.

rxm7706:
> I'd prefer skipping this. Conda-forge already dropped it. We will eventually start feeling the pain of not having other Python 3.7 packages not get updated. Easier to let it go.

@jakirkham @h-vetinari 
Apologies for even asking and putting the maintainers through this discussion. 

Would it be possible to support Py37 for a bit, its on us for being behind on our python upgrades for multiple reasons. But so far PyArrow has provided us a great path to get off many legacy vendor / 3rd party integrations. 

Two big reason we were hoping for PyArrow >=10.0 upgrades while using python 3.7 was to leverage all the work w.r.t JDBC drivers and OpenTelemetry Integration.

https://arrow.apache.org/blog/2022/11/01/arrow-flight-sql-jdbc/
https://github.com/apache/arrow/pull/11920

Thank you.

jakirkham:
The challenge that we tend to see as maintainers is folks will raise issues about incompatibilities between different packages. For example in issue ( https://github.com/conda-forge/arrow-cpp-feedstock/issues/695 ) a user raised an issue about an incompatibility between Arrow and Tensorflow. More of these kinds of incompatibilities can creep in when continuing to maintain a legacy Python version for one package and not others. This in turn invites requests to readd the legacy Python version in more places. It's easier to stop this scope creep and increased maintenance burden by simply being clear with users about what is supported/not after making a decision.

Would also add conda-forge is far slower than other parts of the ecosystem to drop Python versions. Consider for example [NEP 29]( https://numpy.org/neps/nep-0029-deprecation_policy.html ), which is followed by many projects, encouraged dropping Python 3.7 ~1yr ago. Whereas conda-forge dropped Python 3.7 near the end of October ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3573 ) (even when some folks were interested in us dropping it quicker ( https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/2623 )).

h-vetinari:
> We will eventually start feeling the pain of not having other Python 3.7 packages not get updated.

It's mostly @rxm7706 that will feel this, but I checked, and the host deps here are not very migration-exposed.

I tend to agree with you, but given that it's essentially no effort (not even an increase in CI jobs), I think it's fine to help out for such an explicit request (though with no promises or guarantees for the future, as I noted in cbc.yaml).

h-vetinari:
> I certainly don't feel strongly enough about this to push against sustained opposition

@jakirkham, can you qualify how strongly you feel about this somewhere between don't-like-it, opposed, and strongly opposed? Could you live with an exception here? Otherwise I'll close this PR.

h-vetinari:
> @jakirkham, can you qualify how strongly you feel about this somewhere between don't-like-it, opposed, and strongly opposed? Could you live with an exception here? Otherwise I'll close this PR.

@jakirkham Could you please comment?

h-vetinari:
sorry, this is one's an ex-parrot...

PR #899, opened 2022-12-04
Reinstate aarch64+CUDA; add ppc64le+CUDA

h-vetinari:
A left-over from #875, this is blocked on the slow-moving[^1] https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210 (more discussion on the why cross-compilation is a de facto necessity [here](https://github.com/conda-forge/arrow-cpp-feedstock/pull/866#issuecomment-1293002068)).

Closes #859 
Closes #659

[^1]: This might be further impacted by nvidia deps [moving](https://github.com/conda-forge/staged-recipes/issues/21382) to conda-forge


h-vetinari:
After https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210 got solved, this is now blocked on https://github.com/conda-forge/nvcc-feedstock/issues/95

h-vetinari:
I'm rebasing this, but need to wait til https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3624 makes it through the CDN

h-vetinari:
It seems this is looking for the wrong gnu stubs?
```
# include <gnu/stubs-32.h>
          ^~~~~~~~~~~~~~~~
1 error generated.
```

AFAICT they are named [differently](https://github.com/conda-forge/linux-sysroot-feedstock/blob/main/recipe/meta.yaml#L131-L132) in our linux-sysroot? Probably needs a patch?

h-vetinari:
Thanks a lot for the help here Isuru! Checking out where the test failures are coming from in #1015. Sidenote, I saw that arrow wants to find libevent, got an open [PR](https://github.com/conda-forge/libevent-feedstock/pull/24) to make that happen.

h-vetinari:
An interesting error while testing pyarrow on the new CUDA builds...

aarch:
```
export SRC_DIR=/home/conda/feedstock_root/build_artifacts/apache-arrow_1681520079529/test_tmp
$SRC_DIR/conda_test_runner.sh: $PREFIX/bin/python: /lib/ld-linux-aarch64.so.1: bad ELF interpreter: No such file or directory
Tests failed for pyarrow-11.0.0-py38h220f70f_14_cuda.conda - moving package to /home/conda/feedstock_root/build_artifacts/broken
```
ppc:
```
export SRC_DIR=/home/conda/feedstock_root/build_artifacts/apache-arrow_1681520086495/test_tmp
$SRC_DIR/conda_test_runner.sh: $PREFIX/bin/python: /lib64/ld64.so.2: bad ELF interpreter: No such file or directory
Tests failed for pyarrow-11.0.0-py311h6081924_14_cuda.conda - moving package to /home/conda/feedstock_root/build_artifacts/broken
```

h-vetinari:
> An interesting error while testing pyarrow on the new CUDA builds...

The reason this is weird to me is that the error appears on `bin/python` invocation, and doesn't seem to have anything to do with arrow. It sounds sysroot-related, but I don't know these bits well enough to understand what's going on yet.

h-vetinari:
> > An interesting error while testing pyarrow on the new CUDA builds...
> 
> The reason this is weird to me is that the error appears on `bin/python` invocation, and doesn't seem to have anything to do with arrow. It sounds sysroot-related, but I don't know these bits well enough to understand what's going on yet.

So after some investigation[^1], `ld-linux-aarch64.so.1` is [there](https://github.com/regro/libcfgraph/blob/master/artifacts/sysroot_linux-aarch64/conda-forge/noarch/sysroot_linux-aarch64-2.17-h43d7e78_13.json#L223) under `lib64`, which is also confirmed by the link check.
```
INFO (pyarrow,lib/python3.11/site-packages/pyarrow/_cuda.cpython-311-aarch64-linux-gnu.so): Needed DSO aarch64-conda-linux-gnu/sysroot/lib64/ld-linux-aarch64.so.1 found in CDT/compiler package conda-forge::sysroot_linux-aarch64-2.17-h43d7e78_13
```

This raises the question why we need the sysroot at runtime (I mean yes, we need the dynamic linker, but `ld_impl_linux-aarch64` is there, which doesn't [have](https://github.com/regro/libcfgraph/blob/master/artifacts/ld_impl_linux-aarch64/conda-forge/linux-aarch64/ld_impl_linux-aarch64-2.40-h2d8c526_0.json#L236-L241) the lib though, presumably because it shouldn't be needed just to run the dynamic linker).

[^1]: first thought it's an ABI mismatch, cross-compile problem, or a `/lib` vs `/lib64` issue, but I think now that those are less likely.

h-vetinari:
Nope. Still same error. 😑

h-vetinari:
@isuruf, I'm a bit stuck on what's causing the `ld` / ELF mismatch here. Would you have an idea what it could be?

isuruf:
@conda-forge-admin, rerender

h-vetinari:
Do you want to temporarily disable the tests here so there's a more easily-downloadable artefact?

isuruf:
Should be fixed by https://github.com/conda-forge/docker-images/pull/233

h-vetinari:
Something with $CUDA_HOME (or so?) seems to not yet be set up 100% correctly:
```
import: 'pyarrow.cuda'
  File "/home/conda/feedstock_root/build_artifacts/apache-arrow_1681808893494/test_tmp/run_test.py", line 29, in <module>
    import pyarrow.cuda
  File "$PREFIX/lib/python3.9/site-packages/pyarrow/cuda.py", line 21, in <module>
    from pyarrow._cuda import (Context, IpcMemHandle, CudaBuffer,
ImportError: libcuda.so.1: cannot open shared object file: No such file or directory
```

h-vetinari:
Should have been fixed by https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/234

h-vetinari:
## Cross-compiling CUDA ✅🥳 

This PR is now ready for primetime, and I couldn't be more thrilled that we finally got this working. 🤩 

<details>
<summary>Context</summary>

Before arrow 10.x, each python version had a separate CI job, which was already regularly running into the 6h time limit when running in emulation (I was constantly restarting these jobs after merging to the `9.0.x` and `8.0.x` branches, taking up to [13](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=650407&view=results)(!) runs to pass). This was also why we never added support for CUDA-on-ppc, because it would have made that restarting pain many times worse still.

With [arrow 10](https://github.com/conda-forge/arrow-cpp-feedstock/pull/875), `libarrow` finally became independent of the python version; allowing us to build it only once, and collapse the huge build matrix[^1] to do all python versions in one job per arch. That was a big benefit, but made it impossible to compile CUDA support in emulation, so we ended up dropping support for CUDA-on-aarch at the time, pending cross-compilation capabilities.

</details>

### IOW, we can now add back CUDA-on-aarch, and add support for CUDA-on-ppc!

I've also prepared backports of this PR for all our maintenance branches ([10.0.x](https://github.com/conda-forge/arrow-cpp-feedstock/pull/1017), [9.0.x](https://github.com/conda-forge/arrow-cpp-feedstock/pull/1018), [8.0.x](https://github.com/conda-forge/arrow-cpp-feedstock/pull/1019)). Since we had dropped CUDA-on-aarch for arrow 10, backporting there should be obvious, and since cross-compilation hugely cuts down our CI times (and gets rid of the necessity to restart!) on the `9.0.x` and `8.0.x` branches, it should be a no-brainer there as well[^2]. In order to get green CI, those PRs also fix https://github.com/conda-forge/arrow-cpp-feedstock/issues/1016 on all branches.

PTAL @conda-forge/arrow-cpp 
CC @conda-forge/core (just for the good news 😊)

[^1]: 72 jobs at a time when we still had to build for both OpenSSL 1.1.1 & 3.0; would have been 80 had we included CUDA-on-ppc.

[^2]: An argument could be made for leaving the "feature" of CUDA-on-ppc out of the maintenance branches, but since it costs ~nothing and gets us consistency across all branches, I think we should do it.

PS: It took quite a bit of effort to get this working (wouldn't have happened without @isuruf; a huge thanks! 🙏), here's some indication of what was necessary (no claim of exhaustiveness; just what I'm aware of)...
* https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3624
* https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210
* https://github.com/conda-forge/docker-images/pull/229
* https://github.com/conda-forge/nvcc-feedstock/pull/96
* https://github.com/conda-forge/docker-images/pull/233
* https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/234

h-vetinari:
Seems people are less excited about this than I am? 🤔 😅 
(but then, I guess it's me benefitting the most from not having to restart CI all the time)

Unless there are comments to the contrary, I'll merge this (and the maintenance PRs) in 24h.

recipe/meta.yaml
```diff
@@ -24,19 +24,20 @@ source:
       - patches/0001-don-t-bake-non-relocatable-CMAKE_INSTALL_FULL_LIBDIR.patch
       # backport apache/arrow#34019 to disable useless pkgconfig search that takes ~15min
       - patches/0002-GH-33882-C-Don-t-find-.pc-files-with-ARROW_BUILD_STA.patch
+      # backport apache/arrow#34498, 34878 & #34881 for pandas 2.0 compatibility
+      - patches/0003-GH-34404-Python-Failing-tests-because-pandas.Index-c.patch
+      - patches/0004-GH-15070-Python-CI-Compatibility-with-pandas-2.0-348.patch
+      - patches/0005-GH-34880-Python-CI-Fix-Windows-tests-failing-with-la.patch
```
kkraus14, line 30: Is there a reason we need to pull these into this PR? Could this be done in a separate PR?

recipe/build-arrow.sh
```diff
@@ -42,13 +42,19 @@ then
...
 fi
 
-if [[ "${target_platform}" == "osx-arm64" ]]; then
-    EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DCLANG_EXECUTABLE=${BUILD_PREFIX}/bin/clang -DLLVM_LINK_EXECUTABLE=${BUILD_PREFIX}/bin/llvm-link"
+if [[ "${build_platform}" != "${target_platform}" ]]; then
+    # point to a usable protoc/grpc_cpp_plugin if we're cross-compiling
+    EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DProtobuf_PROTOC_EXECUTABLE=$BUILD_PREFIX/bin/protoc"
+    if [[ ! -f ${BUILD_PREFIX}/bin/${CONDA_TOOLCHAIN_HOST}-clang ]]; then
+        ln -sf ${BUILD_PREFIX}/bin/clang ${BUILD_PREFIX}/bin/${CONDA_TOOLCHAIN_HOST}-clang
+    fi
```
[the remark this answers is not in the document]
h-vetinari: This was Isuru's change, not sure what problem he (fore)saw.

recipe/meta.yaml
h-vetinari: Yes, CI is failing without it. For the maintenance branches I just added `pandas <2` as a test constraint, but I wanted to have at least one live version that's fully compatible with pandas 2.0.

h-vetinari: I also don't see the point of separating things into different PRs - we already have a huge amount of (re)builds on this feedstock, so I try to condense them where reasonable. However, the commits are reasonably-separated and if we want we can easily back out stuff later.

(Both these things are in stark contrast to how upstream development works, but the same development model doesn't apply well to the feedstock IMO).

kkraus14: Apologies, I didn't read the recipe close enough where I assumed there was a `pandas<2` constraint that was being removed alongside the patches. Given there isn't one this seems much more fine to me.

kkraus14 (approved):

recipe/meta.yaml
h-vetinari: No worries, thanks for the review! :)

recipe/build-arrow.sh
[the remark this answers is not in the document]
isuruf: File exists on macos, but not linux.

PR #947, opened 2023-01-30
activate.sh: Perform read permission check before symlinking

AndreasAlbertQC:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)

resolves #946 

@h-vetinari let me know if any other change propagation is needed, I'm not very familiar with the branching setup in this repo.

PR #948, opened 2023-01-30
fix changed mimalloc / substrait hash

h-vetinari:
#947 failed after merging due to 
```
-- SHA256 hash of
    /Users/runner/miniforge3/conda-bld/apache-arrow_1675118156023/work/cpp/build/mimalloc_ep-prefix/src/v2.0.6.tar.gz
  does not match expected value
    expected: '9f05c94cc2b017ed13698834ac2a3567b6339a8bde27640df5a1581d49d05ce5'
      actual: '57f86a6bc0045e590358eb2baf1b288edaa383388f0263360444e54a7ac3a2a7'
-- Hash mismatch, removing...
CMake Error at mimalloc_ep-stamp/download-mimalloc_ep.cmake:170 (message):
  Each download failed!

    error: downloading 'https://apache.jfrog.io/artifactory/arrow/thirdparty/7.0.0/mimalloc-v2.0.6.tar.gz' failed
```

Seems someone changed the artifacts...


recipe/meta.yaml
```diff
@@ -27,6 +27,10 @@ source:
       # with the usual relocation treatement and leads to other issues, see #935;
       # replace with placeholder that allows interested users to fix the file path
       - patches/0003-don-t-bake-non-relocatable-CMAKE_INSTALL_FULL_LIBDIR.patch
+      # apparently, someone changed the artifacts under
+      #   * https://github.com/substrait-io/substrait/archive/v0.6.0.tar.gz
+      #   * https://apache.jfrog.io/artifactory/arrow/thirdparty/7.0.0/mimalloc-v2.0.6.tar.gz
+      - patches/0004-fix-changed-thirdparty-artifact-hashes.patch
```
h-vetinari, line 33: @kou @kkraus14 @xhochy 

Do we need to be concerned that the hashes for the versioned tarballs of substrait and mimalloc changed? This happened sometime in the last couple of hours - I assume it's benign, but in the absolute worst case it could indicate a compromised artifact.

@conda-forge/arrow-cpp

kkraus14: I'm guessing this is related to https://github.com/community/community/discussions/45830

h-vetinari:
After github rolled back their tarball changes, this is now passing again, so the PR is not necessary anymore.

Issue #962, opened 2023-02-09
Is it necessary to install cudatoolkit with pyarrow 11 on Linux?


Hoxbro:
### Comment:

When installing pyarrow 11 on Linux cudatoolkit is installed. It is a pretty big dependency:

``` bash
mamba create -n tmp python=3.10 pyarrow=11 --dry-run --offline 2>&1 | grep cuda
  + cudatoolkit               11.8.0  h37601d7_11          conda-forge/linux-64      667MB
```

This is not downloaded with pyarrow=10.

When installing the environment I can see this is because of ucx:

``` python
mamba repoquery whoneeds -t cudatoolkit

cudatoolkit[11.8.0]
  └─ ucx[1.12.1]
     └─ libarrow[11.0.0]
        ├─ arrow-cpp[11.0.0]
        │  └─ parquet-cpp[1.5.1]
        │     └─ pyarrow[11.0.0]
        └─ pyarrow already visited

```

h-vetinari:
ucx indeed got added for arrow 11 (on the feedstock, haven't looked at backporting this yet), though I agree that cudatoolkit is a bit heavier than expected.

Are you using the CPU or CUDA-builds for arrow? I guess we could restrict it to the CUDA builds.

That said, there's a larger theme here that arrow keeps growing non-trivial dependencies. I guess we could introduce a separate output for a "minimal" arrow (`libarrow-core`?). Problem is that the layering is not obvious, i.e. a lot of things get compiled into the same shared library, and it's not clear that we could just separate things without building fully independent outputs.

CC @conda-forge/arrow-cpp

jorisvandenbossche:
I just ran into the same issue. This now gets installed for a normal CPU build (I was using a simple `mamba install pyarrow` on ubuntu). 

> I guess we could restrict it to the CUDA builds.

Is it on the arrow-cpp side that you would not depend on UCX for CPU builds, or it is that `ucx` itself should not depend on cudatoolkit for CPU builds?

It seems that ucx already has some logic around this: https://github.com/conda-forge/ucx-split-feedstock/blob/dd05d5af3a6c4902093ecff980564b6605633124/recipe/meta.yaml#L59-L60 (maybe I am interpreting it wrongly, but it seems it should already not depend on cudatoolkit if `cuda_compiler_version` is "None")



[...]

PR #963, opened 2023-02-16
Remove ucx dependency for now

jorisvandenbossche:
See https://github.com/conda-forge/arrow-cpp-feedstock/issues/962#issuecomment-1433172084

Checklist

### DIFF
--- a/.ci_support/linux_64_cuda_compiler_version10.2.yaml
+++ b/.ci_support/linux_64_cuda_compiler_version10.2.yaml
@@ -68,8 +68,6 @@ target_platform:
 - linux-64
 thrift_cpp:
 - 0.16.0
-ucx:
-- 1.12.1
 zip_keys:
 - - c_compiler_version
   - cxx_compiler_version
--- a/.ci_support/linux_64_cuda_compiler_versionNone.yaml
+++ b/.ci_support/linux_64_cuda_compiler_versionNone.yaml
@@ -68,8 +68,6 @@ target_platform:
 - linux-64
 thrift_cpp:
 - 0.16.0
-ucx:
-- 1.12.1
 zip_keys:
 - - c_compiler_version
   - cxx_compiler_version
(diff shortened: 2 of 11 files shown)


h-vetinari:
Thanks for the PR. You'll have to adapt build.sh as well (it might work without adaptation but it would be confusing). The commit history resp. git blame should be relatively self-explanatory I hope. :)

jorisvandenbossche:

jakirkham:
cc @kkraus14

jakirkham:
I'm hoping we can solve this on the `ucx` end instead of needing this change. There is a 1.14.0 release coming up, which would be a good time to coordinate package changes.

We already have a change for `libnuma` in the queue ( https://github.com/conda-forge/ucx-split-feedstock/issues/112 ).

For example issue ( https://github.com/conda-forge/ucx-split-feedstock/issues/66 ) proposes folding the CPU & GPU builds into one build and then making `cudatoolkit` a [`run_constrained`]( https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-constrained ) dependency (so optional) instead of `run` (required), which drop the need for the selector and drop pulling in `cudatoolkit` (by default)

kkraus14:
I agree that I'd much prefer to fix the UCX package than disable it in the libarrow builds if possible, but understand if there's a short term fix we need to make here in the meantime.

jorisvandenbossche:
Yes, this PR is only meant for "until the ucx issues is solved". But to me it seems quite unsure what would be the exact timeline for getting those fixes in for ucx 1.14 (also just being unfamiliar with the package, I can't judge that), while I would consider the pyarrow issue as a serious regression in user experience (hugely blowing up every environment with pyarrow on linux), something we should fix on the short-term IMO.

jorisvandenbossche:
The timeout on aarch64, is that a known issue? Or could someone restart that build?

h-vetinari:
Let's undo this regression (even though somewhat subjective), and then readd it once ucx is ready.

h-vetinari:
FYI: now that https://github.com/conda-forge/ucx-split-feedstock/pull/111 has landed, I added a PR to re-enable ucx support in https://github.com/conda-forge/arrow-cpp-feedstock/pull/1011

PR #1033, opened 2023-04-29
Make activation script more robust

### LINKED ISSUE #1032: Permission denied when recreating symlink

### Issue

There seems to be an issue with the [`activate.sh`](https://github.com/conda-forge/arrow-cpp-feedstock/blob/main/recipe/activate.sh) script failing in case the user doesn't have permissions to modify the symlink.

For example, in the following Dockerfile, if the environment is activated by `root`, then all subsequent non-root user activations fail with

```
ln: failed to create symbolic link '/opt/conda/share/gdb/auto-load//opt/conda/lib/*.py': Permission denied 
```

```Dockerfile
FROM mambaorg/micromamba:git-3208378

RUN micromamba install -y -n base -c conda-forge libarrow

ARG MAMBA_DOCKERFILE_ACTIVATE=1

USER root

RUN echo "Environment containing libarrow activated as root"

USER $MAMBA_USER

RUN echo "Activate as user without root permission"
```

I'm not sure how to best mitigate this, but one simple idea would be to not recreate the symlink if it already exists and points to the correct location.

### Installed packages

```shell
See Dockerfile
```

### Environment info

```shell
See Dockerfile
```

maresb:
Closes #1032

Checklist
* [X] Bumped the build number (if the version is unchanged)
* [X] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

### DIFF
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -10,18 +10,18 @@
 # where the GDB wrappers get installed
 GDB_PREFIX="$CONDA_PREFIX/share/gdb/auto-load"
 
-# If the directory is not writable, nothing can be done
-if [ ! -w "$GDB_PREFIX" ]; then
-    return
-fi
-
 # this needs to be in sync with the respective patch
 PLACEHOLDER="replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX"
 # the paths here are intentionally stacked, see #935, resp.
 # https://github.com/apache/arrow/blob/master/docs/source/cpp/gdb.rst#manual-loading
 WRAPPER_DIR="$GDB_PREFIX/$CONDA_PREFIX/lib"
 
-mkdir -p "$WRAPPER_DIR"
+mkdir -p "$WRAPPER_DIR" || true
+# If the directory is not writable, nothing can be done
+if [ ! -w "$WRAPPER_DIR" ]; then
+    return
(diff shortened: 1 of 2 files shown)


h-vetinari (approved):
Thanks for the PR. I find the quoting a tad excessive and the scenario (of having rights to the folder but not to the symlink) a bit weird, but fine by me.

maresb:
Thanks for the feedback @h-vetinari.

I agree that the particular scenario of rights to the folder but not the symlink is a little weird. I did it out of an abundance of caution. However, the issue in #1032 should be solved by only ensuring that the directory is writable because the symlink and directory should be created simultaneously by the same user. I could switch from `rm; ln -s` back to `ln -sf` if you prefer.

While some of the quotes (like those around `$f`) aren't necessary, isn't it good practice in general to quote all Bash variables? For instance, the published version fails in case the prefix contains a space.

h-vetinari:
> While some of the quotes (like those around `$f`) aren't necessary

I was mainly talking about those, because we know the lib names don't contain spaces. The quoted paths are fine.

PR #1048, opened 2023-05-05
Build with opentelemetry support (on unix)

h-vetinari:


h-vetinari:
Looks like either opentelemetry or arrow [need](https://stackoverflow.com/questions/17184547/getting-undeclared-identifier-errors-in-winsock-h) an `#include <winsock2.h>` or `#include <ws2tcpip.h>` somewhere

xhochy:
@h-vetinari You might get better error messages if you deactivate the unity builds. The error message triggers a gut feeling of "oh, maybe there is a macro defined twice".

xhochy:
The `UnwrapScan` function is probably missing an `ARROW_EXPORT` declaration: https://github.com/apache/arrow/blob/cdefbb8f4b4183b29fbcdb014af1f6fc0030475c/cpp/src/arrow/util/tracing_internal.h#LL125C1-L125C1

Although it is internal, it is used across library boundaries and thus it needs to be exported. Thus should be addressed upstream.

h-vetinari:
Good analysis @xhochy. The build now succeeded, but we run into something else when trying to load the library:
```
import: 'pyarrow'
[libprotobuf ERROR D:\bld\libprotobuf-split_1670986007159\work\src\google\protobuf\descriptor_database.cc:642] File already exists in database: opentelemetry/proto/common/v1/common.proto
[libprotobuf FATAL D:\bld\libprotobuf-split_1670986007159\work\src\google\protobuf\descriptor.cc:1986] CHECK failed: GeneratedDatabase()->Add(encoded_file_descriptor, size): 
  File "D:\bld\apache-arrow_1683937204901\test_tmp\run_test.py", line 2, in <module>
    import pyarrow
  File "D:\bld\apache-arrow_1683937204901\_test_env\lib\site-packages\pyarrow\__init__.py", line 65, in <module>
    import pyarrow.lib as _lib
ImportError: DLL load failed while importing lib: A dynamic link library (DLL) initialization routine failed.
```

I haven't seen this before, but it seems there's some double-loading(?) of proto information into a protobuf-internal database?

If so, it sounds like arrow shouldn't be re-loading the otel proto definitions?

CC @lidavidm

h-vetinari:
> If so, it sounds like arrow shouldn't be re-loading the otel proto definitions?

To top things off, it looks like there's a third potential source for this (at least in conda-forge), in addition to opentelemetry & arrow: https://github.com/conda-forge/proto-opentelemetry-proto-feedstock

The opentelemetry lib [includes](https://github.com/conda-forge/cpp-opentelemetry-sdk-feedstock/blob/main/recipe/meta.yaml#L35) this as a build dependency(?), which seems wrong somehow? Shouldn't it be a host-dependence?

Also, if there's a need to match the proto-opentelemetry-proto version here, I think this should become a run-export of libopentelemetry-cpp, rather than duplicating the version ("0.19") here and then trying to keep it in sync?

xhochy:
> File already exists in database: opentelemetry/proto/common/v1/common.proto
2023-05-15T06:23:18.8525852Z [libprotobuf FATAL

This happens if two libraries load the same protobuf definition in a different, conflicting version. We have the same problem with all the onnx-* packages. The workaround for this is build all (except at most one) with static protobuf so that each library gets its own protobuf (global) namespace.

h-vetinari:
Thanks for the input!

> This happens if two libraries load the same protobuf definition in a different, conflicting version. 

Can we not ensure that they get the same protobuf definitions? There's even a specific [feedstock](https://github.com/conda-forge/proto-opentelemetry-proto-feedstock) for those.

h-vetinari:
BTW: The recent commits also showed that the unity build was directly implicated in the windows failures. Switching it back on lead to failure again. Should I raise an issue for this upstream?

lidavidm:
> The opentelemetry lib [includes](https://github.com/conda-forge/cpp-opentelemetry-sdk-feedstock/blob/main/recipe/meta.yaml#L35) this as a build dependency(?), which seems wrong somehow? Shouldn't it be a host-dependence?

The Protobuf definitions are only needed to generate code, not at runtime. But the version needs to be matched, so if there's a better way...

> This happens if two libraries load the same protobuf definition in a different, conflicting version. We have the same problem with all the onnx-* packages. The workaround for this is build all (except at most one) with static protobuf so that each library gets its own protobuf (global) namespace.

Hmm, something seems wrong. Arrow itself shouldn't be loading the OpenTelemetry Protobuf generated code. Either we actually built a bundled copy of OpenTelemetry (doesn't seem so) or we still statically linked OpenTelemetry somehow? (Shouldn't be possible?) Or possibly multiple OpenTelemetry libraries included the Protobuf generated code?

> BTW: The recent commits also showed that the unity build was directly implicated in the windows failures. Switching it back on lead to failure again. Should I raise an issue for this upstream?

Yes, please do.

h-vetinari:
Alrighty, close to two years later, we're finally there! 🥳 

(well, except windows, as always...)

PR #1058, opened 2023-05-15
Enable test suite of `libarrow`

h-vetinari:
Since #875, we're now running with the test suite for `pyarrow`. While working on #1048, I wanted to ensure that also the library portion works as intended, so thought about switching on the test suite for `libarrow`.

After some initial attempts it looks like this is failing, hence will need some iteration together with upstream.


h-vetinari:
Alrighty, after over 2 years of working towards this, we now have the C++ tests running for `libarrow` 🥳

PR #1065, opened 2023-05-29
Rework libarrow activate.sh with finer-grained checks and logging

### LINKED ISSUE #1060: Activate.sh can't run in parallel

### Issue

I can't reproduce it, but it "feels like" if multiple activations of the same environment happens in parallel (on NFS), then the activation fails with:
```
ln: failed to create symbolic link '/.../share/gdb/auto-load//.../lib/libarrow.so.1200.0.0-gdb.py': File exists
```

### Installed packages

```shell
libarrow-12.0.0-h6564b11_2_cpu.conda
```

### Environment info

```shell
shell level : 2
conda version : 22.11.1
conda-build version : 3.21.9
python version : 3.10.8.final.0
```

maresb:
I expect this to fix #1060.

Checklist

### DIFF
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -24,9 +24,11 @@ _la_placeholder="replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX
 # https://github.com/apache/arrow/blob/master/docs/source/cpp/gdb.rst#manual-loading
 _la_wrapper_dir="$_la_gdb_prefix/$CONDA_PREFIX/lib"
 
-_la_log "   _la_gdb_prefix: $_la_gdb_prefix"
-_la_log "  _la_placeholder: $_la_placeholder"
-_la_log "  _la_wrapper_dir: $_la_wrapper_dir"
+_la_log "          _la_gdb_prefix: $_la_gdb_prefix"
+_la_log "         _la_placeholder: $_la_placeholder"
+_la_log "  _la_target_wrapper_dir: $_la_target_wrapper_dir"
+_la_log "  _la_actual_wrapper_dir: $_la_gdb_prefix/$_la_placeholder/lib"
+_la_log "  content of that folder:\n$(ls -l $_la_gdb_prefix/$_la_placeholder/lib)"
 
 # there's only one lib in the _la_placeholder folder, but the libname changes
 # based on the version so use a loop instead of hardcoding it.
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -19,13 +19,13 @@ _la_log "Beginning libarrow activation."
 _la_gdb_prefix="$CONDA_PREFIX/share/gdb/auto-load"
 
 # this needs to be in sync with ARROW_GDB_INSTALL_DIR in build.sh
-PLACEHOLDER="replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX"
+_la_placeholder="replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX"
 # the paths here are intentionally stacked, see #935, resp.
 # https://github.com/apache/arrow/blob/master/docs/source/cpp/gdb.rst#manual-loading
 WRAPPER_DIR="$_la_gdb_prefix/$CONDA_PREFIX/lib"
 
 _la_log "   _la_gdb_prefix: $_la_gdb_prefix"
-_la_log "  PLACEHOLDER: $PLACEHOLDER"
+_la_log "  _la_placeholder: $_la_placeholder"
 _la_log "  WRAPPER_DIR: $WRAPPER_DIR"
 
 mkdir -p "$WRAPPER_DIR" || true
@@ -35,9 +35,9 @@ if [ ! -w "$WRAPPER_DIR" ]; then
     return
 fi
 
-# there's only one lib in the placeholder folder, but the libname changes
+# there's only one lib in the _la_placeholder folder, but the libname changes
 # based on the version so use a loop instead of hardcoding it.
-for target in "$_la_gdb_prefix/$PLACEHOLDER/lib/"*.py; do
+for target in "$_la_gdb_prefix/$_la_placeholder/lib/"*.py; do
     if [ ! -e "$target" ]; then
         # If the file doesn't exist, skip this iteration of the loop.
         # (This happens when no files are found, in which case the
@@ -60,3 +60,4 @@ _la_log "Libarrow activation complete."
 
 unset _la_log
 unset _la_gdb_prefix
+unset _la_placeholder
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -46,9 +46,7 @@ for target in "$_la_gdb_prefix/$_la_placeholder/lib/"*.py; do
         continue
     fi
     symlink="$_la_wrapper_dir/$(basename "$target")"
-    # Check if symbolic link already exists and points to correct file
     if [ -L "$symlink" ] && [ "$(readlink "$symlink")" = "$target" ]; then
-        # Stop if it does
         _la_log "Symlink '$symlink' already exists and points to '$target', skipping."
         continue
     fi
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -37,7 +37,7 @@ for _la_target in "$_la_gdb_prefix/$_la_placeholder/lib/"*.py; do
         # If the file doesn't exist, skip this iteration of the loop.
         # (This happens when no files are found, in which case the
         # loop runs with target equal to the pattern itself.)
-        _la_log "Folder $_la_gdb_prefix/$_la_placeholder/lib seems to not contain *.py files, skipping"
+        _la_log "Folder '$_la_gdb_prefix/$_la_placeholder/lib' seems to not contain .py files, skipping"
         continue
     fi
     _la_symlink="$_la_wrapper_dir/$(basename "$_la_target")"
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -9,7 +9,11 @@
 
 _la_log() {
     if [ "$CF_LIBARROW_ACTIVATE_LOGGING" = "1" ]; then
-        echo "${BASH_SOURCE[0]} DEBUG: $*"
+        # The following loop is necessary to handle multi-line strings
+        # like for the output of `ls -al`.
+        while IFS= read -r line; do
+            echo "${BASH_SOURCE[0]} DEBUG: $line"
+        done <<< "$*"
     fi
 }
 
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -7,6 +7,14 @@
 # doesn't come with a deactivate script, because the symlink
 # is benign and doesn't need to be deleted.
 
+_la_log() {
+    if [ "$LIBARROW_ACTIVATE_LOGGING" = "1" ]; then
+        echo "DEBUG ${BASH_SOURCE[0]}: $*"
+    fi
+}
+
+_la_log "Beginning libarrow activation."
+
 # where the GDB wrappers get installed
 GDB_PREFIX="$CONDA_PREFIX/share/gdb/auto-load"
 
@@ -16,9 +24,14 @@ PLACEHOLDER="replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX"
 # https://github.com/apache/arrow/blob/master/docs/source/cpp/gdb.rst#manual-loading
 WRAPPER_DIR="$GDB_PREFIX/$CONDA_PREFIX/lib"
 
+_la_log "   GDB_PREFIX: $GDB_PREFIX"
+_la_log "  PLACEHOLDER: $PLACEHOLDER"
+_la_log "  WRAPPER_DIR: $WRAPPER_DIR"
+
 mkdir -p "$WRAPPER_DIR" || true
 # If the directory is not writable, nothing can be done
 if [ ! -w "$WRAPPER_DIR" ]; then
+    _la_log "Wrapper directory '$WRAPPER_DIR' is not writable, aborting."
     return
 fi
 
@@ -29,13 +42,20 @@ for target in "$GDB_PREFIX/$PLACEHOLDER/lib/"*.py; do
         # If the file doesn't exist, skip this iteration of the loop.
         # (This happens when no files are found, in which case the
         # loop runs with target equal to the pattern itself.)
+        _la_log "Skipping extraneous iteration with target as match pattern '$target'"
         continue
     fi
     symlink="$WRAPPER_DIR/$(basename "$target")"
     # Check if symbolic link already exists and points to correct file
     if [ -L "$symlink" ] && [ "$(readlink "$symlink")" = "$target" ]; then
         # Stop if it does
+        _la_log "Symlink '$symlink' already exists and points to '$target', skipping."
         continue
     fi
+    _la_log "Creating symlink '$symlink' pointing to '$target'"
     ln -sf "$target" "$symlink"
 done
+
+_la_log "Libarrow activation complete."
+
+unset _la_log
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -28,13 +28,6 @@ _la_log "   _la_gdb_prefix: $_la_gdb_prefix"
 _la_log "  _la_placeholder: $_la_placeholder"
 _la_log "  _la_wrapper_dir: $_la_wrapper_dir"
 
-mkdir -p "$_la_wrapper_dir" || true
-# If the directory is not writable, nothing can be done
-if [ ! -w "$_la_wrapper_dir" ]; then
-    _la_log "Wrapper directory '$_la_wrapper_dir' is not writable, aborting."
-    return
-fi
-
 # there's only one lib in the _la_placeholder folder, but the libname changes
 # based on the version so use a loop instead of hardcoding it.
 for target in "$_la_gdb_prefix/$_la_placeholder/lib/"*.py; do
@@ -51,6 +44,12 @@ for target in "$_la_gdb_prefix/$_la_placeholder/lib/"*.py; do
         continue
     fi
     _la_log "Creating symlink '$symlink' pointing to '$target'"
+    mkdir -p "$_la_wrapper_dir" || true
+    # If the directory is not writable, nothing can be done
+    if [ ! -w "$_la_wrapper_dir" ]; then
+        _la_log "Wrapper directory '$_la_wrapper_dir' is not writable, aborting."
+        continue
+    fi
     ln -sf "$target" "$symlink"
 done
 
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -45,15 +45,12 @@ for _la_target in "$_la_gdb_prefix/$_la_placeholder/lib/"*.py; do
     fi
     _la_log "Creating symlink '$_la_symlink' pointing to '$_la_target'"
     mkdir -p "$_la_wrapper_dir" || true
-    # If the directory is not writable, nothing can be done
-    if [ ! -w "$_la_wrapper_dir" ]; then
-        echo -n "${BASH_SOURCE[0]} ERROR: Wrapper directory '$_la_wrapper_dir' is not "
-        echo -n "writable, so we can't create the symlink '$_la_symlink' pointing to "
-        echo -n "'$_la_target'."
+    if ! ln -sf "$_la_target" "$_la_symlink"; then
+        echo -n "${BASH_SOURCE[0]} ERROR: Failed to create symlink from "
+        echo -n "'$_la_target' to '$_la_symlink'"
         echo
         continue
     fi
-    ln -sf "$_la_target" "$_la_symlink"
 done
 
 _la_log "Libarrow activation complete."
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -6,7 +6,7 @@
 
 # doesn't come with a deactivate script, because the symlink
 # is benign and doesn't need to be deleted.
-
+CF_LIBARROW_ACTIVATE_LOGGING=1
 _la_log() {
     if [ "$CF_LIBARROW_ACTIVATE_LOGGING" = "1" ]; then
         # The following loop is necessary to handle multi-line strings
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -37,7 +37,7 @@ _la_log "    _la_orig_install_dir: $_la_orig_install_dir"
 _la_log "  content of that folder:"
 _la_log "$(ls -al "$_la_orig_install_dir" | sed 's/^/      /')"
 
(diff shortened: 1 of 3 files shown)


recipe/activate.sh
```diff
@@ -22,18 +22,34 @@ if [ ! -w "$WRAPPER_DIR" ]; then
...
         continue
     fi
-    target="$WRAPPER_DIR/$(basename "$f")"
-    # We have write permissions to WRAPPER_DIR but not necessarily target.
-    # Thus it's safest to delete the target in case it already exists.
-    rm -f "$target"
-    ln -s "$f" "$WRAPPER_DIR/$(basename "$f")"
+    symlink="$WRAPPER_DIR/$(basename "$target")"
+    # Check if symbolic link already exists and points to correct file
+    if [ -L "$symlink" ] && [ "$(readlink "$symlink")" = "$target" ]; then
```
alippai, line 39: This is the most important part. Covers every activation after the link is created

recipe/activate.sh
```diff
@@ -22,18 +22,34 @@ if [ ! -w "$WRAPPER_DIR" ]; then
...
+        # Stop if it does
+        continue
+    fi
+    # We have write permissions to WRAPPER_DIR but not necessarily the symlink.
+    # Thus it's safest to delete the symlink in case it already exists.
+    rm -f "$symlink"
+    # Check if symlink still exists after trying to delete it
+    if [ -e "$symlink" ]; then
```
alippai, line 47: Instead of `rm + existence check + ln -s` you can create a new unique link with random name and `mv` to the final place to make it atomic. In two parallel processes `rm` + `-e` can happen the same time and the two `ln -s` will still collide.

maresb: I modified it. Is that what you had in mind?

recipe/activate.sh
```diff
@@ -22,18 +22,29 @@ if [ ! -w "$WRAPPER_DIR" ]; then
...
+    symlink="$WRAPPER_DIR/$(basename "$target")"
+    # Check if symbolic link already exists and points to correct file
+    if [ -L "$symlink" ] && [ "$(readlink "$symlink")" = "$target" ]; then
+        # Stop if it does
+        continue
+    fi
+    # We have write permissions to WRAPPER_DIR but not necessarily the symlink.
+    # Thus rather than editing the symlink, we should create a temporary symlink
+    # and then move it into place.
+    temp_symlink="$(mktemp --tmpdir gdb-symlink-for-arrow-cpp-XXXXX)"
```
h-vetinari, line 46: I had to look up what this does, and it's IMO not obvious (at first I thought it's a directory because of `--tmpdir`).

Also, it seems that OSX doesn't support that option:
```
mktemp: illegal option -- -
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix
```

Why not just `-p .`?

recipe/activate.sh
```diff
@@ -22,18 +22,29 @@ if [ ! -w "$WRAPPER_DIR" ]; then
...
+    ln -sf "$target" "$temp_symlink"
+    mv "$temp_symlink" "$symlink"
 done
+
```
h-vetinari, line 50: extraneous newline

h-vetinari (commented):
Thanks for the PR! Some comments.

recipe/activate.sh
alippai: I wonder if simply calling `ln -f` does the same. Overwrites anything there atomically and changes the owner etc.

maresb:
Thanks a lot @h-vetinari and @alippai for the detailed comments!

I reworked a lot of the script. Highlights:
* I switched back to using `ln -sf` to create the symlink, as it seems to do the right thing. (I think my changing this in #1033 was a mistake.)
* I'm no longer checking the permissions of the directory. We only attempt to create the link if necessary, so if we are going to fail, then fail on `ln -sf`.
* If we do fail, then it's a problem, and so print a message about it, including the activation script filename so that the user has a shot at debugging.
* Add a logging function so that I can set `LIBARROW_ACTIVATE_LOGGING=1` and see what's going on.
* Clean up all these environment variables after we're done with them.

recipe/activate.sh
```diff
@@ -7,33 +7,57 @@
...
+        continue
+    fi
+    _la_symlink="$_la_wrapper_dir/$(basename "$_la_target")"
+    if [ -L "$_la_symlink" ] && [ "$(readlink "$_la_symlink")" = "$_la_target" ]; then
+        _la_log "symlink '$_la_symlink' already exists and points to '$_la_target', skipping."
+        continue
+    fi
+    _la_log "Creating symlink '$_la_symlink' pointing to '$_la_target'"
+    mkdir -p "$_la_wrapper_dir" || true
+    if ! ln -sf "$_la_target" "$_la_symlink"; then
```
h-vetinari, line 48: I'm not a fan of tests with side effects. In this case I can see that it's justified, but please add a comment.

```suggestion
    # this check also creates the symlink; if it fails, we enter the if-branch.
    if ! ln -sf "$_la_target" "$_la_symlink"; then
```

recipe/activate.sh
```diff
@@ -7,33 +7,57 @@
 # doesn't come with a deactivate script, because the symlink
 # is benign and doesn't need to be deleted.
 
+_la_log() {
+    if [ "$LIBARROW_ACTIVATE_LOGGING" = "1" ]; then
+        echo "${BASH_SOURCE[0]} DEBUG: $*"
+    fi
```
h-vetinari, line 13: We don't get to "steal" arrows namespace here, let's rename the variable to something like:
```suggestion
    if [ "$CF_LIBARROW_ACTIVATE_LOGGING" = "1" ]; then
        echo "${BASH_SOURCE[0]} DEBUG: $*"
    fi
```

recipe/activate.sh
```diff
@@ -7,33 +7,57 @@
...
 # based on the version so use a loop instead of hardcoding it.
-for f in "$GDB_PREFIX/$PLACEHOLDER/lib/"*.py; do
-    if [ ! -e "$f" ]; then
+for _la_target in "$_la_gdb_prefix/$_la_placeholder/lib/"*.py; do
+    if [ ! -e "$_la_target" ]; then
         # If the file doesn't exist, skip this iteration of the loop.
         # (This happens when no files are found, in which case the
-        # loop runs with f equal to the pattern itself.)
+        # loop runs with target equal to the pattern itself.)
+        _la_log "Skipping extraneous iteration with target as match pattern '$_la_target'"
```
h-vetinari, line 38: That log message is not very clear (unless you also read the comment in the activation script). Perhaps:
```suggestion
        _la_log "Folder $_la_gdb_prefix/$_la_placeholder/lib seems to not contain *.py files, skipping"
```

recipe/activate.sh
```diff
@@ -7,33 +7,57 @@
...
+_la_wrapper_dir="$_la_gdb_prefix/$CONDA_PREFIX/lib"
 
-mkdir -p "$WRAPPER_DIR" || true
-# If the directory is not writable, nothing can be done
-if [ ! -w "$WRAPPER_DIR" ]; then
-    return
-fi
+_la_log "   _la_gdb_prefix: $_la_gdb_prefix"
+_la_log "  _la_placeholder: $_la_placeholder"
+_la_log "  _la_wrapper_dir: $_la_wrapper_dir"
```
h-vetinari, line 29: If we're infodumping, might want to show which files are actually there (implies renaming of `_la_wrapper_dir`, if you like that approach).
```suggestion
_la_log "          _la_gdb_prefix: $_la_gdb_prefix"
_la_log "         _la_placeholder: $_la_placeholder"
_la_log "  _la_target_wrapper_dir: $_la_target_wrapper_dir"
_la_log "  _la_actual_wrapper_dir: $_la_gdb_prefix/$_la_placeholder/lib"
_la_log "  content of that folder:\n$(ls -l $_la_gdb_prefix/$_la_placeholder/lib)"
```

recipe/activate.sh
```diff
@@ -7,33 +7,57 @@
...
 done
+
+_la_log "Libarrow activation complete."
+
+unset _la_log
+unset _la_gdb_prefix
+unset _la_placeholder
+unset _la_wrapper_dir
+unset _la_symlink
+unset _la_target
```
h-vetinari, line 63: Please sort these alphabetically

h-vetinari (changes requested):
16 commits is a bit hefty for updating the activate script. If you'd like me not to squash-merge, please clean up the history to ~5 commits tops (e.g. variable privatization, add logging, update the logic, or some permutation thereof)

maresb:
Great suggestions @h-vetinari, thanks a lot!!!

I think I addressed your points.

Probably this is best done as a squash-merge, thanks for asking.

maresb:
Oops, one more thing...

recipe/activate.sh
```diff
@@ -7,33 +7,60 @@
 # doesn't come with a deactivate script, because the symlink
 # is benign and doesn't need to be deleted.
 
+_la_log() {
+    if [ "$CF_LIBARROW_ACTIVATE_LOGGING" = "1" ]; then
```
h-vetinari, line 11: Can you temporarily make this unconditional? I'd like to see a CI run through with the logging activated. Then we can revert that commit and merge.

Also, shouldn't this be `==`?

recipe/activate.sh
```diff
@@ -7,33 +7,60 @@
...
-GDB_PREFIX="$CONDA_PREFIX/share/gdb/auto-load"
+_la_gdb_prefix="$CONDA_PREFIX/share/gdb/auto-load"
 
 # this needs to be in sync with ARROW_GDB_INSTALL_DIR in build.sh
-PLACEHOLDER="replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX"
+_la_placeholder="replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX"
 # the paths here are intentionally stacked, see #935, resp.
 # https://github.com/apache/arrow/blob/master/docs/source/cpp/gdb.rst#manual-loading
-WRAPPER_DIR="$GDB_PREFIX/$CONDA_PREFIX/lib"
+_la_wrapper_dir="$_la_gdb_prefix/$CONDA_PREFIX/lib"
```
h-vetinari, line 25: There's a couple places that now either need to be renamed `_la_wrapper_dir` -> `_la_target_wrapper_dir`

recipe/activate.sh
```diff
@@ -7,33 +7,60 @@
...
 
-mkdir -p "$WRAPPER_DIR" || true
-# If the directory is not writable, nothing can be done
-if [ ! -w "$WRAPPER_DIR" ]; then
-    return
-fi
+_la_log "          _la_gdb_prefix: $_la_gdb_prefix"
+_la_log "         _la_placeholder: $_la_placeholder"
+_la_log "  _la_target_wrapper_dir: $_la_target_wrapper_dir"
```
h-vetinari, line 29: Alternatively, you could just do this:
```suggestion
_la_log "  _la_target_wrapper_dir: $_la_wrapper_dir"
```

maresb:
@h-vetinari, why do you think I should be using `==` in `"$CF_LIBARROW_ACTIVATE_LOGGING" = "1"`?

Apart from that, I think I addressed all your new points.

I'm a bit confused by the output while testing. I'm testing using the following Dockerfile:
```Dockerfile
FROM mambaorg/micromamba:git-3208378
RUN micromamba install -y -n base -c conda-forge libarrow nano
COPY activate.sh /opt/conda/etc/conda/activate.d/libarrow_activate.sh
ENV CF_LIBARROW_ACTIVATE_LOGGING=1
USER root
```
Running `docker build -t debug-libarrow . && docker run --rm -it debug-libarrow` produces the following output:

```bash
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG: Beginning libarrow activation.
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:           _la_gdb_prefix: /opt/conda/share/gdb/auto-load
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:          _la_placeholder: replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:   _la_actual_wrapper_dir: /opt/conda/share/gdb/auto-load//opt/conda/lib
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:   _la_target_wrapper_dir: /opt/conda/share/gdb/auto-load/replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX/lib
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:   content of that folder:
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:       total 12
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:       drwxr-xr-x 2 mambauser mambauser 4096 May 30 07:29 .
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:       drwxr-xr-x 3 mambauser mambauser 4096 May 30 07:29 ..
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:       -rw-rw-r-- 1 mambauser mambauser  953 May 30 07:29 libarrow.so.1200.0.0-gdb.py
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG: Folder $_la_target_wrapper_dir seems to not contain .py files, skipping
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG: Libarrow activation complete.
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG: Beginning libarrow activation.
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:           _la_gdb_prefix: /opt/conda/share/gdb/auto-load
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:          _la_placeholder: replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:   _la_actual_wrapper_dir: /opt/conda/share/gdb/auto-load//opt/conda/lib
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:   _la_target_wrapper_dir: /opt/conda/share/gdb/auto-load/replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX/lib
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:   content of that folder:
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:       total 12
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:       drwxr-xr-x 2 mambauser mambauser 4096 May 30 07:29 .
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:       drwxr-xr-x 3 mambauser mambauser 4096 May 30 07:29 ..
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:       -rw-rw-r-- 1 mambauser mambauser  953 May 30 07:29 libarrow.so.1200.0.0-gdb.py
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:    _la_target: /opt/conda/share/gdb/auto-load/replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX/lib/libarrow.so.1200.0.0-gdb.py
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG:   _la_symlink: /opt/conda/share/gdb/auto-load//opt/conda/lib/libarrow.so.1200.0.0-gdb.py
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG: Creating symlink $_la_symlink pointing to $_la_target
/opt/conda/etc/conda/activate.d/libarrow_activate.sh DEBUG: Libarrow activation complete.
```

Namely the activation script seems to be called twice. The first time the wrapper directory appears to contain the `.py` file, but in the for loop, `_la_target` is set to the pattern. I'm really confused. Any idea what's going on here?

h-vetinari:
> @h-vetinari, why do you think I should be using `==` in `"$CF_LIBARROW_ACTIVATE_LOGGING" = "1"`?

I'm used to this syntax:
```
if [[ "$PKG_NAME" == "scipy" ]]; then
```
Perhaps with single brackets comes single equality signs? It's confusing 😅 

> Namely the activation script seems to be called twice. The first time the wrapper directory appears to contain the `.py` file, but in the for loop, `_la_target` is set to the pattern. I'm really confused. Any idea what's going on here?

To be honest, I don't know. The activation script is only installed for one package, so it _should_ normally only be activated once.

> ```
> DEBUG:   _la_actual_wrapper_dir: /opt/conda/share/gdb/auto-load//opt/conda/lib
> DEBUG:   _la_target_wrapper_dir: /opt/conda/share/gdb/auto-load/replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX/lib
> ```

So perhaps this was me using suboptimal naming (given that for a symlink, "target" means something else), but to me those two folders are named incorrectly. The actual, physically-on-the-filestystem gdb wrapper is in `/opt/conda/share/gdb/auto-load/replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX/lib` (what I meant by `_la_actual_wrapper_dir`). Where we want to _have_ a wrapper (you can choose another name than `target`) is `/opt/conda/share/gdb/auto-load//opt/conda/lib`

h-vetinari:
> The activation script is only installed for one package, so it _should_ normally only be activated once.

Let's just see what the CI does

recipe/activate.sh
```diff
@@ -6,34 +6,70 @@
...
-fi
+_la_log "          _la_gdb_prefix: $_la_gdb_prefix"
+_la_log "         _la_placeholder: $_la_placeholder"
+_la_log "         _la_symlink_dir: $_la_symlink_dir"
+_la_log "    _la_orig_install_dir: $_la_orig_install_dir"
+_la_log "  content of that folder:"
+_la_log "$(ls -al "$_la_orig_install_dir" | sed 's/^/      /')"
 
-# there's only one lib in that folder, but the libname changes
+# there's only one lib in the _la_placeholder folder, but the libname changes
```
h-vetinari, line 39: Can be done after we remove/revert `CF_LIBARROW_ACTIVATE_LOGGING=1` again.
```suggestion
# there's only one lib in the _la_orig_install_dir folder, but the libname changes
```

maresb: Got it. Updated the [todo list](https://github.com/conda-forge/arrow-cpp-feedstock/pull/1065#issuecomment-1568210670).

maresb: Should we also rerender?

h-vetinari: No, this feedstock was very recently rerendered, and I don't like squashing your changes with those of the bot.

recipe/activate.sh
```diff
@@ -6,34 +6,70 @@
 
 # doesn't come with a deactivate script, because the symlink
 # is benign and doesn't need to be deleted.
+CF_LIBARROW_ACTIVATE_LOGGING=1
+_la_log() {
+    if [ "$CF_LIBARROW_ACTIVATE_LOGGING" = "1" ]; then
+        # The following loop is necessary to handle multi-line strings
+        # like for the output of `ls -al`.
+        while IFS= read -r line; do
+            echo "${BASH_SOURCE[0]} DEBUG: $line"
```
xhochy, line 15: While the script implies bash at the top, I think this is also executed with other shells like `zsh` and thus this will break them.

maresb: Oh yikes, thanks for pointing this out.

Would this solve it?

```diff
     if [ "$CF_LIBARROW_ACTIVATE_LOGGING" = "1" ]; then
         # The following loop is necessary to handle multi-line strings
         # like for the output of `ls -al`.
-        while IFS= read -r line; do
-            echo "${BASH_SOURCE[0]} DEBUG: $line"
-        done <<< "$*"
+        printf '%s\n' "$*" | while IFS= read -r line
+        do
+            echo "$CONDA_PREFIX/etc/conda/activate.d/libarrow_activate.sh DEBUG: $line"
+        done
     fi
 }
```

xhochy: Yes, this will work.

maresb:
The run completed successfully. Now pushing the changes to complete the [todo](https://github.com/conda-forge/arrow-cpp-feedstock/pull/1065#issuecomment-1568210670)...

h-vetinari (approved):
LGTM (for squash merge)

maresb:
I can't imagine the last batch of commits changing much. Is the following failure of `linux linux_64_cuda_compiler_versionNone` a transient error?

```python
__________________________ test_total_bytes_allocated __________________________

    def test_total_bytes_allocated():
>       assert pa.total_allocated_bytes() == 0
E       assert 1344 == 0
E        +  where 1344 = <built-in function total_allocated_bytes>()
E        +    where <built-in function total_allocated_bytes> = pa.total_allocated_bytes

test_array.py:41: AssertionError
```

h-vetinari:
> The run completed successfully. Now pushing the changes to complete the [todo](https://github.com/conda-forge/arrow-cpp-feedstock/pull/1065#issuecomment-1568210670)...

This really wouldn't have needed a new PR. Please close the other one. Also, I _still_ don't see a 

> Is the following failure of `linux linux_64_cuda_compiler_versionNone` a transient error?

Yes

h-vetinari:
You managed to comprehensively confuse github with two PRs having the exact same HEAD, and now I can't restart the builds. Another reason not to open a new PR unless really necessary. 🙃

maresb:
Sorry about that, I was trying to save time by starting a new CI run before the old one finished.

Would it help anything if I pushed an empty commit to the other branch and open-close-cycled the other PR, or something like that?

h-vetinari:
> Would it help anything if I pushed an empty commit to the other branch and open-close-cycled the other PR, or something like that?

No worries, I already fixed it here. That commit will not show up in the squash merge

maresb:
Looks like one of the runners is a bit messed up.

h-vetinari:
> Looks like one of the runners is a bit messed up.

Yeah, I don't even get the dialogue for restarting things. Mainly I was waiting if @xhochy still wanted to have some input here. But let's take this for a spin. :)

plattenschieber:
We stumbled upon an issue in our ci which seems to be related to this PR, as yesterdays build seems fine and todays doesn't work anymore 😅
```
Transaction finished
/workspace/repo/.env/etc/conda/activate.d/libarrow_activate.sh: line 11: CF_LIBARROW_ACTIVATE_LOGGING: unbound variable
```

The only change I could see was in the build number of `pyarrow` 12.0.0
`libarrow 12.0.0 h96638e8_3_cpu` worked fine -> `libarrow 12.0.0 h96638e8_4_cpu` got error from above

maresb:
Hi @plattenschieber, thanks for the report! Indeed, this is would be a natural consequence of your Bash shell having `set -u` to enable unbound variable checking. We should set a default value to prevent this error in this case. I'll submit a PR for it now.

PR #1067, opened 2023-06-01
Fix unbound variable in activate script

maresb:
We were checking `CF_LIBARROW_ACTIVATE_LOGGING` without a default value. This led to failure under `set -u`. For original report from @plattenschieber, see <https://github.com/conda-forge/arrow-cpp-feedstock/pull/1065#issuecomment-1571699535>:

> We stumbled upon an issue in our ci which seems to be related to this PR, as yesterdays build seems fine and todays doesn't work anymore :sweat_smile:
> 
> ```
> Transaction finished
> /workspace/repo/.env/etc/conda/activate.d/libarrow_activate.sh: line 11: CF_LIBARROW_ACTIVATE_LOGGING: unbound variable
> ```
> 
> The only change I could see was in the build number of `pyarrow` 12.0.0 `libarrow 12.0.0 h96638e8_3_cpu` worked fine -> `libarrow 12.0.0 h96638e8_4_cpu` got error from above

Checklist
* [X] Bumped the build number (if the version is unchanged)

### DIFF
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -8,7 +8,7 @@
 # is benign and doesn't need to be deleted.
 
 _la_log() {
-    if [ "$CF_LIBARROW_ACTIVATE_LOGGING" = "1" ]; then
+    if [ "${CF_LIBARROW_ACTIVATE_LOGGING:-}" = "1" ]; then
         # The following loop is necessary to handle multi-line strings
         # like for the output of `ls -al`.
         printf '%s\n' "$*" | while IFS= read -r line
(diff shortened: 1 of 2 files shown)


h-vetinari:
Ok, this makes sense, and the linux-64 runs are green already. Thanks!

maresb:
From a systematic viewpoint, maybe the CI Bash shell should `set -u` so that we catch this sort of thing ahead of time?

h-vetinari:
> From a systematic viewpoint, maybe the CI Bash shell should `set -u` so that we catch this sort of thing ahead of time?

To catch this in the activation scripts would need a global change in conda-forge (IIUC), and that sounds like it would break _a lot_ of recipes for what amounts to a cosmetic improvement. I doubt it's worth the effort.

PR #1068, opened 2023-06-02
Stabilize CI by skipping crashy/flaky tests

h-vetinari:
Test runs on aarch and ppc have been really crashy recently. Turns out they all point to the exact same tests (even though the error shown in azure implies otherwise):
```
test_memory.py ..............FEEEEE                                      [ 36%]
```

Let's just skip those. Also added a few tests that fail occasionally, but regularly enough to be annoying. There's also the occasional error on windows where something like `with tempfile.TemporaryDirectory() as tempdir2:` can fail upon resource destruction with:
```
E    PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\tmpdw6ktuts\\x\\part-0.arrow'
```
These however happen pretty much anywhere, and while very annoying, aren't really worth wholesale skipping these tests over.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -367,6 +367,8 @@ outputs:
         # skip tests that raise SIGINT and crash the test suite
         {% set tests_to_skip = tests_to_skip + " or (test_csv and test_cancellation)" %}  # [linux]
         {% set tests_to_skip = tests_to_skip + " or (test_flight and test_interrupt)" %}  # [linux]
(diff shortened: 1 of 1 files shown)


h-vetinari:
Ah, the irony of _still_ running into
```
This job was abandoned. We have detected that logs from the agent may have not finished uploading. We have included our in-memory record of all log lines uploaded before we lost contact with the agent:
```
in this PR. 🤣

PR #1082, opened 2023-06-07
[main] Rebuild for protobuf423

h-vetinari:
Migrator stuck on pytorch, which we only use as a test dependency anyway.


h-vetinari:
Joining a long line of baffling behaviours around CUDA-on-aarch, the persistent failures (with or without the specific skips from #1081), the problems seem to go away with protobuf 4. I just don't care enough about CUDA+aarch to spend yet more hours on investigating/fixing why this is blowing through all the diskspace, but if anyone want to debug and has a PR that passes CI several times in a row, I'm happy to merge it.

PR #1101, opened 2023-07-01
[main] Rebuild for libprotobuf4233_libgrpc156

h-vetinari:
Migrator is stuck on test deps, it's IMO not worth the hassle for microscopic amounts of extra test coverage (that never caused a failure either). In particular, even if we do it manually, the bot will not move on to dependents.


h-vetinari:
Ah damn, this will need the google-cloud-cpp PR first...

PR #1116, opened 2023-07-11
Allow 64kb pagesize on aarch64

benfred:
Closes #1115

Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #1119, opened 2023-07-14
Backport #1116 to v9.x

benfred:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)


h-vetinari (approved):
Thank you! :)

varlackc (commented):
The package is missing the `dev_url` and `doc_url` on the `meta.yaml` file.

  https://github.com/conda-forge/arrow-cpp-feedstock/blob/cf8da1ff92dbb03a53d3f56e713357634ab59883/recipe/meta.yaml#L142&L147

  the following might be so valid `dev_url` and `doc_url`

  dev_url: https://github.com/apache/arrow
  doc_url: https://arrow.apache.org

h-vetinari:
> The package is missing the `dev_url` and `doc_url` on the `meta.yaml` file.

What I consider important is that https://github.com/apache/arrow is there, which it is. A `doc_url` IMO only makes sense if it is different from both `home` and `dev_url`. But in any case, you can raise a PR (to `main`) if you want.

jakirkham:
@varlackc could you please raise that as a new issue? That seems unrelated to the changes here

More generally the approach conda-forge has taken is these other URLs are optional. We include them in examples and docs to show how people can use them, but don't require them

Recognize other recipe / package collections may benefit from this extra data. Though in that case would suggest raising issues on feedstocks to discuss with maintainers whether they would be open to including them and (ideally) submitting a PR once agreement has been reached to add those URLs

However asking in unrelated PRs is not ideal. So would request refraining from that in the future. Does that make sense?

PR #1122, opened 2023-07-20
Fix cross-compiled CUDA jobs that are running out of space

### LINKED ISSUE #1114: cross-compiled CUDA builds running out of disk space
I don't know what change caused this (perhaps something in the CUDA setup...), but since about a month or so, cross-compiling CUDA consistently blows through the disk space of the azure workers, failing the job in an often un-restartable way. 

I've tried fixing this in various ways (#1075, #1081, 7c267128a11af2d62855898d2812297fd9ecacc0, a8ca8f793c920dad4dfb2f377a53ab886c8603d3, 555a42cdd11c50a3c7ac5d105248e62d91982525). The problem exists on both ppc & aarch, but for ppc at least, the various fixes seem to have mostly settled things, but for aarch it's still failing 9 times out of 10.

(note, the only reason I disabled aws-sdk-cpp is that jobs started failing again with migrating to a new version that had some more features enabled, and a footprint around 40MB; this is being tackled in https://github.com/conda-forge/google-cloud-cpp-feedstock/issues/141).

CC @conda-forge/cuda-compiler @jakirkham @isuruf

PS. By chance I've seen that qt also has disk space problems, and worked around this by [disabling](https://github.com/conda-forge/qt-main-feedstock/blob/eb39f78c2e2c2c5c67c41b809813a9ea8b1dc028/recipe/build.sh#L66-L70) the use of precompiled headers. Arrow has an option `ARROW_USE_PRECOMPILED_HEADERS`, but it's already off [by default](https://github.com/apache/arrow/blob/apache-arrow-12.0.1/cpp/cmake_modules/DefineOptions.cmake#L176).

h-vetinari:
Closes #1114 with https://github.com/conda-forge/conda-smithy/pull/1747 (after various other approaches to increase space were not enough).

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -8,10 +8,10 @@ jobs:
     vmImage: ubuntu-latest
   strategy:
     matrix:
-      linux_64_cuda_compiler_version10.2:
-        CONFIG: linux_64_cuda_compiler_version10.2
+      linux_64_cuda_compiler_version11.2:
+        CONFIG: linux_64_cuda_compiler_version11.2
         UPLOAD_PACKAGES: 'True'
-        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-cos7-cuda:10.2
+        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-cuda:11.2
       linux_64_cuda_compiler_versionNone:
         CONFIG: linux_64_cuda_compiler_versionNone
         UPLOAD_PACKAGES: 'True'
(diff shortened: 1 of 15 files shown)


h-vetinari:
Interesting... Now it's even running out of space before the main compilation starts at all.

jakirkham:
That doesn't sound good. Let's retry once we have new Docker images

h-vetinari:
Still no dice even with the 2GB reduction from https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/258. 😑 

Really curious where this blow-up is coming from. It already fails during the creation of the build environment now, before we even executed anything from the scripts.

jakirkham:
Tried restarting after PR ( https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/258 ) was merged and this still failed due to the same issue

jakirkham:
Glad to see this working! 🥳 

Thanks Axel! 🙏

PR #1124, opened 2023-07-24
arrow 13.0.0

h-vetinari:
~Let's see what building off of the latest tag brings. This helps prepare for when the bot comes around, but also to button up https://github.com/apache/arrow/pull/35637~

For arrow 13 I'm also planning to remove the old naming of the outputs `arrow-cpp` & `arrow-cpp-proc`.

Closes #1139

### DIFF
--- a/.ci_support/migrations/aws_crt_cpp0230.yaml
+++ b/.ci_support/migrations/aws_crt_cpp0230.yaml
@@ -1,8 +1,0 @@
-__migrator:
-  build_number: 1
-  kind: version
-  migration_number: 1
-  automerge: true
-aws_crt_cpp:
-- 0.23.0
-migrator_ts: 1692537530.8322937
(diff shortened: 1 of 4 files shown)


h-vetinari:
@conda-forge/arrow-cpp, are people OK with dropping the old `arrow-cpp` compatibility outputs?

We've now reached the point where the last 4 releases (10-13) all have `libarrow`, so I think we could/should drop it. I can add a rule to the linter to warn any remaining users of `arrow-cpp` to move to `libarrow.

xhochy:
Result of an offline discussion on `arrow-cpp` between @h-vetinari is that we will keep the output for version 13 and will drop them with 14. Currently, a package that builds against Arrow C++ must use `arrow-cpp` as we don't have the output yet for version 9.

h-vetinari:
> Currently, a package that builds against Arrow C++ [for all currently supported versions] must use `arrow-cpp` as we don't have the output yet for version 9.

... which is solved once we drop off arrow v9 after the v13 migration. 🙃

h-vetinari:
Actually, since this passed already before, in it goes... :)

PR #1127, opened 2023-07-28
[main] bail out of activation if environment is not writable

### LINKED ISSUE #1126: Activating read-only conda environment with arrow installed

### Issue

When we create a conda environment with this package installed in a read-only environment I get the following error when I attempt to activate the environment. 

```shell
mkdir: cannot create directory ‘/home/conda/rag-llm/envs/rag-llm-ai-toolbox/share/gdb/auto-load//home/conda/rag-llm/envs’: Permission denied
ln: failed to create symbolic link '/home/conda/rag-llm/envs/rag-llm-ai-toolbox/share/gdb/auto-load//home/conda/rag-llm/envs/rag-llm-ai-toolbox/lib/libarrow.so.1100.0.0-gdb.py': No such file or directory
/home/conda/rag-llm/envs/rag-llm-ai-toolbox/etc/conda/activate.d/libarrow_activate.sh ERROR: Failed to create symlink from '/home/conda/rag-llm/envs/rag-llm-ai-toolbox/share/gdb/auto-load/replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX/lib/libarrow.so.1100.0.0-gdb.py' to '/home/conda/rag-llm/envs/rag-llm-ai-toolbox/share/gdb/auto-load//home/conda/rag-llm/envs/rag-llm-ai-toolbox/lib/libarrow.so.1100.0.0-gdb.py'
```

### Installed packages

```shell
Not relevant.
```

### Environment info

```shell
Not relevant.
```

h-vetinari:
Reinstate 7c0287a84ba8303b12c97e397f2b24a0239b6b46

Closes #1126

Also pick up some small improvements from #1124

### DIFF
--- a/recipe/activate.sh
+++ b/recipe/activate.sh
@@ -23,6 +23,12 @@ _la_log "Beginning libarrow activation."
 # where the GDB wrappers get installed
 _la_gdb_prefix="$CONDA_PREFIX/share/gdb/auto-load"
 
+# If the directory is not writable, nothing can be done
+if [ ! -w "$_la_gdb_prefix" ]; then
+    _la_log 'No rights to modify $_la_gdb_prefix, cannot create symlink!'
+    return
+fi
+
 # this needs to be in sync with ARROW_GDB_INSTALL_DIR in build.sh
 _la_placeholder="replace_this_section_with_absolute_slashed_path_to_CONDA_PREFIX"
 # the paths here are intentionally stacked, see #935, resp.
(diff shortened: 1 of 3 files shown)


maresb:
What's the purpose of the GDB wrappers in the first place? (Despite having contributed I don't actually understand what's going on. :joy:)

h-vetinari:
> What's the purpose of the GDB wrappers in the first place?

gdb needs to know where to look for symbols, and has a very specific path pattern it expects. Since that pattern depends on the exact path of the environment (which we cannot know at packaging time), the only option is to do something with an activation script (or not at all).

But where that approach is not possible, we might as well drop gdb support. It's not that important.

maresb:
Ok, so it's only if someone wants to use a debugger with libarrow? That doesn't sound so important.

Maybe we should modify the log message to say so?

> If you don't plan to use the GDB debugger with libarrow, then this warning can be safely ignored.

h-vetinari:
> Ok, so it's only if someone wants to use a debugger with libarrow? That doesn't sound so important.

Yes and yes.

> Maybe we should modify the log message to say so?

OK

maresb (approved):
This is great, thanks for this!

Maybe we want to also add something similar to `ERROR: Failed to create symlink...`?

h-vetinari:
I've switched off my laptop already. Perhaps the next time. ;)
Or feel free to submit a PR that switches the error to (visible by default) warning. 🙃

PR #1201, opened 2023-10-14
arrow 14.0.0 with new setup (more granular libraries; no more legacy outputs)

### LINKED ISSUE #1035: Package some form of "minimal" libarrow?

h-vetinari:
Combines #1170 with a slimmed-down version of #1175

Closes #1035
Closes #1170 
Closes #1175
Closes #1218


h-vetinari:
Some review notes:
* the removal of the compat outputs was the plan we took in #1124, to give people some more time to adapt. We'll also be able to introduce a linter-rule for anyone still using `arrow-cpp`.
* gandiva now depends on libllvm
* with @raulcd doing the lion's share, we finally tackled the split of libarrow into more granular components. The only big chunk left to shave off is the dependence on the cloud provider bindings, but that's not ready yet
* the manual distribution of files in `install-libarrow.*` is obviously not pretty and could in principle be largely replaced by conda's `files:` directive, however, it probably doesn't work for _everything_ (e.g. activation scripts and handling the symlink that gets created by `libarrow` during the build of `libarrow-*`).
* the (only) minio-server available in conda-forge is newer than what arrow supports, and so we need to hack the fixtures a little. This only concerns the test suite and should otherwise be unproblematic (before it didn't get tested at all!).

h-vetinari:
@conda-forge/arrow-cpp, would be happy for reviews or feedback here :)

xhochy (approved):
There are probably ways to speed up the build a bit but from the content-perspective, this PR looks good to me.

h-vetinari:
Thanks for the review! :)

jorisvandenbossche:
@h-vetinari thanks a lot for finishing this splitting!

As a follow-up, should we add a `pyarrow-base` or `pyarrow-core` (I think both suffixes are being used) that only depends on `libarrow`?

raulcd:
I was planning on trying to work on this during the next couple of weeks now that this has been merged and I have released 14.0.1.

h-vetinari:
> As a follow-up, should we add a `pyarrow-base` or `pyarrow-core` (I think both suffixes are being used) that only depends on `libarrow`?

Based on discussion in https://github.com/apache/arrow/issues/38536 and the work towards that which went into arrow 14 already, I think we should definitely try!

> I was planning on trying to work on this during the next couple of weeks now that this has been merged and I have released 14.0.1.

Great!

PR #1255, opened 2023-11-27
Create subsets of pyarrow package with pyarrow-core < pyarrow < pyarrow-all and update to Arrow v16.0.0

raulcd:
Checklist

As discussed in https://github.com/conda-forge/arrow-cpp-feedstock/pull/1201 this PR tries to add a new `pyarrow-base` that only depends on `libarrow` and `libparquet`.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -23,13 +23,15 @@ source:
       - patches/0002-Revert-GH-37803-CI-Dev-Python-Release-and-merge-scri.patch
       # backport https://github.com/apache/arrow/pull/40230 for glog 0.7 compat
       - patches/0003-GH-40181-C-Support-glog-0.7-build-40230.patch
+      # backport https://github.com/apache/arrow/pull/40412 to plug a python memory leak
+      - patches/0004-GH-37989-Plug-reference-leaks.patch
   # testing-submodule not part of release tarball
   - git_url: https://github.com/apache/arrow-testing.git
     git_rev: ad82a736c170e97b7c8c035ebd8a801c17eec170
     folder: testing
 
 build:
-  number: 4
+  number: 5
   # for cuda support, building with one version is enough to be compatible with
   # all later versions, since arrow is only using libcuda, and not libcudart.
   skip: true  # [cuda_compiler_version not in ("None", cuda_compiler_version_min)]
--- a/recipe/patches/0004-GH-37989-Plug-reference-leaks.patch
+++ b/recipe/patches/0004-GH-37989-Plug-reference-leaks.patch
@@ -1,0 +1,156 @@
+From defac0d1caff437ad87c6ce646c2f69c3bbfed78 Mon Sep 17 00:00:00 2001
+From: Chun Yang <Chuck.Yang@gmail.com>
+Date: Fri, 15 Mar 2024 08:33:00 -0700
+Subject: [PATCH] GH-37989: [Python] Plug reference leaks when creating Arrow
+ array from Python list of dicts (#40412)
+
+### Rationale for this change
+
+When creating Arrow arrays using `pa.array` from lists of dicts, memory usage is observed to increase over time despite the created arrays going out of scope. The issue appears to only happen for lists of dicts, as opposed to lists of numpy arrays or other types.
+
+### What changes are included in this PR?
+
+This PR makes two changes to _python_to_arrow.cc_, to ensure that new references created by [`PyDict_Items`](https://docs.python.org/3/c-api/dict.html#c.PyDict_Items) and [`PySequence_GetItem`](https://docs.python.org/3/c-api/sequence.html#c.PySequence_GetItem) are properly reference counted via `OwnedRef`.
+
+### Are these changes tested?
+
+The change was tested against the following reproduction script:
+```python
+"""Repro memory increase observed when creating pyarrow arrays."""
+
+# System imports
+import logging
+
+# Third-party imports
+import numpy as np
+import psutil
+import pyarrow as pa
+
+LIST_LENGTH = 5 * (2**20)
+LOGGER = logging.getLogger(__name__)
+
+def initialize_logging() -> None:
+    logging.basicConfig(
+        format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+        level=logging.INFO,
+    )
+
+def get_rss_in_mib() -> float:
+    """Return the Resident Set Size of the current process in MiB."""
+    return psutil.Process().memory_info().rss / 1024 / 1024
+
+def main() -> None:
+    initialize_logging()
+
+    for idx in range(100):
+        data = np.random.randint(256, size=(LIST_LENGTH,), dtype=np.uint8)
+        # data = "a" * LIST_LENGTH
+        pa.array([{"data": data}])
+        if (idx + 1) % 10 == 0:
+            LOGGER.info(
+                "%d dict arrays created, RSS: %.2f MiB", idx + 1, get_rss_in_mib()
+            )
+
+    LOGGER.info("---------")
+
+    for idx in range(100):
+        pa.array(
+            [
+                np.random.randint(256, size=(LIST_LENGTH,), dtype=np.uint8).tobytes(),
+            ]
+        )
+        if (idx + 1) % 10 == 0:
+            LOGGER.info(
+                "%d non-dict arrays created, RSS: %.2f MiB", idx + 1, get_rss_in_mib()
+            )
+
+if __name__ == "__main__":
+    main()
+```
+
+Prior to this change, the reproduction script produces the following output:
+```
+2024-03-07 23:14:17,560 - __main__ - INFO - 10 dict arrays created, RSS: 121.05 MiB
+2024-03-07 23:14:17,698 - __main__ - INFO - 20 dict arrays created, RSS: 171.07 MiB
+2024-03-07 23:14:17,835 - __main__ - INFO - 30 dict arrays created, RSS: 221.09 MiB
+2024-03-07 23:14:17,971 - __main__ - INFO - 40 dict arrays created, RSS: 271.11 MiB
+2024-03-07 23:14:18,109 - __main__ - INFO - 50 dict arrays created, RSS: 320.86 MiB
+2024-03-07 23:14:18,245 - __main__ - INFO - 60 dict arrays created, RSS: 371.65 MiB
+2024-03-07 23:14:18,380 - __main__ - INFO - 70 dict arrays created, RSS: 422.18 MiB
+2024-03-07 23:14:18,516 - __main__ - INFO - 80 dict arrays created, RSS: 472.20 MiB
+2024-03-07 23:14:18,650 - __main__ - INFO - 90 dict arrays created, RSS: 522.21 MiB
+2024-03-07 23:14:18,788 - __main__ - INFO - 100 dict arrays created, RSS: 572.23 MiB
+2024-03-07 23:14:18,789 - __main__ - INFO - ---------
+2024-03-07 23:14:19,001 - __main__ - INFO - 10 non-dict arrays created, RSS: 567.61 MiB
+2024-03-07 23:14:19,211 - __main__ - INFO - 20 non-dict arrays created, RSS: 567.61 MiB
+2024-03-07 23:14:19,417 - __main__ - INFO - 30 non-dict arrays created, RSS: 567.61 MiB
+2024-03-07 23:14:19,623 - __main__ - INFO - 40 non-dict arrays created, RSS: 567.61 MiB
+2024-03-07 23:14:19,832 - __main__ - INFO - 50 non-dict arrays created, RSS: 567.61 MiB
+2024-03-07 23:14:20,047 - __main__ - INFO - 60 non-dict arrays created, RSS: 567.61 MiB
+2024-03-07 23:14:20,253 - __main__ - INFO - 70 non-dict arrays created, RSS: 567.61 MiB
+2024-03-07 23:14:20,499 - __main__ - INFO - 80 non-dict arrays created, RSS: 567.61 MiB
+2024-03-07 23:14:20,725 - __main__ - INFO - 90 non-dict arrays created, RSS: 567.61 MiB
+2024-03-07 23:14:20,950 - __main__ - INFO - 100 non-dict arrays created, RSS: 567.61 MiB
+```
+
+After this change, the output changes to the following. Notice that the Resident Set Size (RSS) no longer increases as more Arrow arrays are created from list of dict.
+```
+2024-03-07 23:14:47,246 - __main__ - INFO - 10 dict arrays created, RSS: 81.73 MiB
+2024-03-07 23:14:47,353 - __main__ - INFO - 20 dict arrays created, RSS: 76.53 MiB
+2024-03-07 23:14:47,445 - __main__ - INFO - 30 dict arrays created, RSS: 82.20 MiB
+2024-03-07 23:14:47,537 - __main__ - INFO - 40 dict arrays created, RSS: 86.59 MiB
+2024-03-07 23:14:47,634 - __main__ - INFO - 50 dict arrays created, RSS: 80.28 MiB
+2024-03-07 23:14:47,734 - __main__ - INFO - 60 dict arrays created, RSS: 85.44 MiB
+2024-03-07 23:14:47,827 - __main__ - INFO - 70 dict arrays created, RSS: 85.44 MiB
+2024-03-07 23:14:47,921 - __main__ - INFO - 80 dict arrays created, RSS: 85.44 MiB
+2024-03-07 23:14:48,024 - __main__ - INFO - 90 dict arrays created, RSS: 82.94 MiB
+2024-03-07 23:14:48,132 - __main__ - INFO - 100 dict arrays created, RSS: 87.84 MiB
+2024-03-07 23:14:48,132 - __main__ - INFO - ---------
+2024-03-07 23:14:48,229 - __main__ - INFO - 10 non-dict arrays created, RSS: 87.84 MiB
+2024-03-07 23:14:48,324 - __main__ - INFO - 20 non-dict arrays created, RSS: 87.84 MiB
+2024-03-07 23:14:48,420 - __main__ - INFO - 30 non-dict arrays created, RSS: 87.84 MiB
+2024-03-07 23:14:48,516 - __main__ - INFO - 40 non-dict arrays created, RSS: 87.84 MiB
+2024-03-07 23:14:48,613 - __main__ - INFO - 50 non-dict arrays created, RSS: 87.84 MiB
+2024-03-07 23:14:48,710 - __main__ - INFO - 60 non-dict arrays created, RSS: 87.84 MiB
+2024-03-07 23:14:48,806 - __main__ - INFO - 70 non-dict arrays created, RSS: 87.84 MiB
+2024-03-07 23:14:48,905 - __main__ - INFO - 80 non-dict arrays created, RSS: 87.84 MiB
+2024-03-07 23:14:49,009 - __main__ - INFO - 90 non-dict arrays created, RSS: 87.84 MiB
+2024-03-07 23:14:49,108 - __main__ - INFO - 100 non-dict arrays created, RSS: 87.84 MiB
+```
+
+When this change is tested against the reproduction script provided in https://github.com/apache/arrow/issues/37989#issue-1924129600, the reported memory increase is no longer observed.
+
+I have not added a unit test, but it may be possible to add one similar to the reproduction scripts used above, provided there's an accurate way to capture process memory usage on all the platforms that Arrow supports, and provided memory usage is not affected by concurrently running tests. If this code could be tested under valgrind, that may be an even better way to go.
+
+### Are there any user-facing changes?
+
+* GitHub Issue: #37989
+
+Authored-by: Chuck Yang <chuck.yang@getcruise.com>
+Signed-off-by: Joris Van den Bossche <jorisvandenbossche@gmail.com>
+---
+ python/pyarrow/src/arrow/python/python_to_arrow.cc | 4 +++-
+ 1 file changed, 3 insertions(+), 1 deletion(-)
+
+diff --git a/python/pyarrow/src/arrow/python/python_to_arrow.cc b/python/pyarrow/src/arrow/python/python_to_arrow.cc
+index d1d94ac17a13e..5da3b561b11a9 100644
+--- a/python/pyarrow/src/arrow/python/python_to_arrow.cc
++++ b/python/pyarrow/src/arrow/python/python_to_arrow.cc
+@@ -1041,7 +1041,8 @@ class PyStructConverter : public StructConverter<PyConverter, PyConverterTrait>
+       case KeyKind::BYTES:
+         return AppendDict(dict, bytes_field_names_.obj());
+       default:
+-        RETURN_NOT_OK(InferKeyKind(PyDict_Items(dict)));
++        OwnedRef item_ref(PyDict_Items(dict));
++        RETURN_NOT_OK(InferKeyKind(item_ref.obj()));
+         if (key_kind_ == KeyKind::UNKNOWN) {
+           // was unable to infer the type which means that all keys are absent
+           return AppendEmpty();
+@@ -1087,6 +1088,7 @@ class PyStructConverter : public StructConverter<PyConverter, PyConverterTrait>
+   Result<std::pair<PyObject*, PyObject*>> GetKeyValuePair(PyObject* seq, int index) {
+     PyObject* pair = PySequence_GetItem(seq, index);
+     RETURN_IF_PYERROR();
++    OwnedRef pair_ref(pair);  // ensure reference count is decreased at scope end
+     if (!PyTuple_Check(pair) || PyTuple_Size(pair) != 2) {
+       return internal::InvalidType(pair, "was expecting tuple of (key, value) pair");
+     }


h-vetinari:
> As discussed in #1201 this PR tries to add a new `pyarrow-base` that only depends on `libarrow` and `libparquet`.

I had thought the idea was to just let `pyarrow` itself depend on `libarrow` and `libparquet`, and provide clear instructions in the error path about how to install the other libraries if necessary (e.g. by installing `libarrow-X` or `libarrow-all`)?

raulcd:
> I had thought the idea was to just let `pyarrow` itself depend on `libarrow` and `libparquet`, and provide clear instructions in the error path about how to install the other libraries if necessary (e.g. by installing `libarrow-X` or `libarrow-all`)?

From the discussions and the conversations I've had with @jorisvandenbossche I thought we wanted to provide both `pyarrow-base` and `pyarrow`.

Currently the error on `pyarrow-base` if we try to import, i.e, dataset is:
```
import: 'pyarrow.dataset'
  File "/home/conda/feedstock_root/build_artifacts/apache-arrow_1701073557860/test_tmp/run_test.py", line 5, in <module>
    import pyarrow.dataset
  File "/home/conda/feedstock_root/build_artifacts/apache-arrow_1701073557860/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/pyarrow/dataset.py", line 58, in <module>
    raise ImportError(
ImportError: The pyarrow installation is not built with support for 'dataset' (libarrow_dataset.so.1400: cannot open shared object file: No such file or directory)
```

I see your point and I am ok with both approaches, probably I would agree that 2 is a better future approach:
1. If we provide both `pyarrow-base` and `pyarrow` we are less disruptive with some of our current users that depend on "extra" libarrow packages.
2. If we provide only `pyarrow` depending only `libarrow` and `libparquet` we have a much cleaner installation going forward but with the current disruption to some of our users.

@jorisvandenbossche what are your thoughts?

recipe/build-pyarrow.bat
```diff
@@ -31,7 +31,7 @@ if "%cuda_compiler_version%"=="None" (
 if %ERRORLEVEL% neq 0 exit 1
 popd
 
-if [%PKG_NAME%] == [pyarrow] (
+if [%PKG_NAME%] == [pyarrow-base] (
```
xhochy, line 34:
```suggestion
if [%PKG_NAME%] NEQ [pyarrow-eests] (
```

recipe/build-pyarrow.sh
```diff
@@ -54,7 +54,7 @@ $PYTHON setup.py \
         install --single-version-externally-managed \
                 --record=record.txt
 
-if [[ "$PKG_NAME" == "pyarrow" ]]; then
+if [[ "$PKG_NAME" == "pyarrow-base" ]]; then
```
xhochy, line 57:
```suggestion
if [[ "$PKG_NAME" != "pyarrow-tests" ]]; then
```

xhochy (changes requested):

recipe/build-pyarrow.bat
raulcd:
```suggestion
if [%PKG_NAME%] NEQ [pyarrow-tests] (
```

jorisvandenbossche:
I haven't yet looked in detail, but some quick drive-by thoughts on the package naming:

- If we want to make (long term, eg also when filesystems can be split off) the most minimal arrow installation possible, shouldn't we also remove the libparquet dependency from the minimal package (whether it is pyarrow or pyarrow-base)?
- If we have a general package that most users will install as the default, and where we want to include parquet, I would personally also include datasets (many features of the `pyarrow.parquet.read_table(..)` actually relies on dataset, like reading a partitioned dataset)
- I think one advantage of keeping a "pyarrow" package that is not the most minimal package is that this is less disrupting _now_ (as mentioned above), but also _later_ if we want to further remove things from "libarrow"/"pyarrow-base".  
  For example when splitting off the compute kernels, that would again be a breaking change if then "pyarrow" wouldn't provide that anymore. Or for example when splitting off the filesystems, that would also break users assuming "pyarrow" comes with S3 support (although in this case there is something to say that this shouldn't be part of base "pyarrow", because then why S3 and not GC or Azure? And we might exactly want to avoid bundling them all in the default install, given you almost never need them all)


A potential alternative is something in between: 

- a `pyarrow-base` that is the absolute minimum (whatever minimum is possible at the moment), and people who rely on this package will more often have to update in case this minimum changes
- a `pyarrow` package that includes some optional components to provide what we consider as a good "default" experience. It's of course debatable what this would be (and also this could change over time), but for example right now I would include parquet and datasets, but not gandiva and flight (because those are used much less, I think, and also have more complex dependencies). And for example when splitting off compute at some point, that could be something that keeps being included in this default `pyarrow` package.
- when someone then wants the features of truly optional pieces like gandiva or flight (and potentially the filesystems in the future), at that point they definitely need to install the additional package explicitly.  
  One question I am wondering here is whether we also want to provide the names "pyarrow-gandiva" and "pyarrow-flight" as aliases of "libarrow-gandiva" etc to make it more user friendly for _python_ users.

h-vetinari:
> some quick drive-by thoughts on the package naming:

I think some degree of breakage is unavoidable if we want to introduce a minimal pyarrow _and_ default to it at some point. The question is whether the error messages that people encounter would be self-explanatory enough to trivially fix things.

> A potential alternative is something in between:

If we really want to have not just `pyarrow-base` and _also_ a `pyarrow` that doesn't contain absolutely everything, then we should have `pyarrow-all` (that would +/- match `libarrow-all`). I'm not 100% excited about this, because `-base` packages are often some packaging work-arounds (like breaking a cyclic build dependence), and not meant to be directly installed.

It wouldn't be the first package with such a setup though, for example [ray](https://github.com/conda-forge/ray-packages-feedstock/blob/main/recipe/meta.yaml) has `ray-core` & `ray-all` as the smallest/largest collection, `ray-default` in between (`ray` was taken), and `ray-*` in for various flavours. Similarly for [modin](https://github.com/conda-forge/modin-feedstock/blob/main/recipe/meta.yaml), which has `modin-core < modin < modin-all` (plus some `modin-*`).

So I guess it'd be reasonable to do `pyarrow-core < pyarrow < pyarrow-all`. Not sure how you'll match that with the PyPI side, but from the POV of conda, we have a lot of design flexibility in this regard.

ZupoLlask:
Hi guys!

So how's this progressing? 👍

raulcd:
@jorisvandenbossche @h-vetinari I have rebased this PR with 15.0.0 so `libparquet` is not required for `pyarrow-core` anymore and it is really only depending on `libarrow`.
At the moment I have created `pyarrow-core` < `pyarrow` < `pyarrow-all`.
- `pyarrow-core` only contains `libarrow` as a run dependency.
- `pyarrow` contains `libarrow`, `libparquet`, `libarrow-acero`, `libarrow-dataset` and `libarrow-substrait` (I am not entirely sure if we should also remove `libarrow-substrait`)
- `pyarrow-all` apart from the `pyarrow` ones also contains as run dependencies `libarrow-flight`, `libarrow-flight-sql` and `libgandiva`.

I think this is good for a review at the moment.
Thanks!

recipe/meta.yaml
```diff
@@ -632,6 +632,20 @@ outputs:
...
+        - '*/libarrow_dataset.so.*'      # [linux]
+        - '*/arrow_dataset.dll'          # [win]
+        - '*/libarrow_substrait.so.*'    # [linux]
+        - '*/arrow_substrait.dll'        # [win]
+        - '*/libarrow_flight.so.*'       # [linux]
+        - '*/arrow_flight.dll'           # [win]
+        - '*/libparquet.so.*'            # [linux]
+        - '*/parquet.dll'                # [win]
+        - '*/libgandiva.so.*'            # [linux]
+        - '*/gandiva.dll'                # [win]
```
h-vetinari, line 648: Here we can actually specify the path (which is much shorter than for site-packages). I also think it'd be more readable to separate linux and win, and AFAICT we need to add osx:
```suggestion
        - 'lib/libarrow_acero.*'              # [unix]
        - 'lib/libarrow_dataset.*'            # [unix]
        - 'lib/libarrow_substrait.*'          # [unix]
        - 'lib/libarrow_flight.*'             # [unix]
        - 'lib/libparquet.*'                  # [unix]
        - 'lib/libgandiva.*'                  # [unix]
        - 'Library/lib/arrow_acero.dll'       # [win]
        - 'Library/lib/arrow_dataset.dll'     # [win]
        - 'Library/lib/arrow_substrait.dll'   # [win]
        - 'Library/lib/arrow_flight.dll'      # [win]
        - 'Library/lib/parquet.dll'           # [win]
        - 'Library/lib/gandiva.dll'           # [win]
```

recipe/meta.yaml
```diff
@@ -644,8 +658,190 @@ outputs:
...
+        - test ! -f ${SP_DIR}/pyarrow/tests/test_array.py                           # [unix]
+        - if exist %SP_DIR%/pyarrow/tests/test_array.py exit 1                      # [win]
+        # Need to remove dot from PY_VER; %MYVAR:x=y% replaces "x" in %MYVAR% with "y"
+        - if not exist %SP_DIR%/pyarrow/_cuda.cp%PY_VER:.=%-win_amd64.pyd exit 1  # [win and cuda_compiler_version != "None"]
+
+        # shared libraries
+        - test -f $PREFIX/lib/libarrow.so                           # [linux]
+        - test -f $PREFIX/lib/libarrow.dylib                        # [osx]
+        - if not exist %LIBRARY_BIN%\arrow.dll exit 1               # [win]
+        - if not exist %LIBRARY_LIB%\arrow.lib exit 1               # [win]
```
h-vetinari, line 724: These tests should be removed, as they're tested in `libarrow`, which is specified to be present at runtime.

recipe/meta.yaml
```diff
@@ -644,8 +658,190 @@ outputs:
...
+        # Expected not included libraries
+        - test ! -f $PREFIX/lib/libarrow_acero.so                   # [unix]
+        - test ! -f $PREFIX/lib/libarrow_dataset.so                 # [unix]
+        - test ! -f $PREFIX/lib/libarrow_flight.so                  # [unix]
+        - test ! -f $PREFIX/lib/libgandiva.so                       # [unix]
+        - test ! -f $PREFIX/lib/libparquet.so                       # [unix]
```
h-vetinari, line 730: This is not useful on osx due to a mismatched extension
```suggestion
        # Expected not included libraries
        - test ! -f $PREFIX/lib/libarrow_acero${SHLIB_EXT}      # [unix]
        - test ! -f $PREFIX/lib/libarrow_dataset${SHLIB_EXT}    # [unix]
        - test ! -f $PREFIX/lib/libarrow_flight${SHLIB_EXT}     # [unix]
        - test ! -f $PREFIX/lib/libgandiva${SHLIB_EXT}          # [unix]
        - test ! -f $PREFIX/lib/libparquet${SHLIB_EXT}          # [unix]
```

recipe/meta.yaml
```diff
@@ -644,8 +658,190 @@ outputs:
...
+        - if not exist %LIBRARY_BIN%\arrow_acero.dll exit 1                         # [win]
+        - if not exist %LIBRARY_LIB%\arrow_acero.lib exit 1                         # [win]
+        - test -f $PREFIX/lib/libarrow_dataset.so                                   # [linux]
+        - test -f $PREFIX/lib/libarrow_dataset.dylib                                # [osx]
+        - if not exist %LIBRARY_BIN%\arrow_dataset.dll exit 1                       # [win]
+        - if not exist %LIBRARY_LIB%\arrow_dataset.lib exit 1                       # [win]
+        - test -f $PREFIX/lib/libparquet.so                                         # [linux]
+        - test -f $PREFIX/lib/libparquet.dylib                                      # [osx]
+        - if not exist %LIBRARY_BIN%\parquet.dll exit 1                             # [win]
+        - if not exist %LIBRARY_LIB%\parquet.lib exit 1                             # [win]
```
h-vetinari, line 823: Remove this

recipe/meta.yaml
```diff
@@ -644,8 +658,190 @@ outputs:
...
+        # Expected not included libraries
+        - test ! -f $PREFIX/lib/libarrow_flight.so                                  # [unix]
+        - test ! -f $PREFIX/lib/libgandiva.so                                       # [unix]
```
h-vetinari, line 826: `${SHLIB_EXT}`

recipe/meta.yaml
```diff
@@ -691,12 +888,15 @@ outputs:
...
-        - if not exist %SP_DIR%\pyarrow\arrow_python_flight.dll exit 1            # [win]
+        - test -f ${SP_DIR}/pyarrow/libarrow_python.so                              # [linux]
+        - test -f ${SP_DIR}/pyarrow/libarrow_python_flight.so                       # [linux]
+        - test -f ${SP_DIR}/pyarrow/libarrow_python_parquet_encryption.so           # [linux]
+        - test -f ${SP_DIR}/pyarrow/libarrow_python.dylib                           # [osx]
+        - test -f ${SP_DIR}/pyarrow/libarrow_python_flight.dylib                    # [osx]
+        - test -f ${SP_DIR}/pyarrow/libarrow_python_parquet_encryption.dylib        # [osx]
+        - if not exist %SP_DIR%\pyarrow\arrow_python.dll exit 1                     # [win]
+        - if not exist %SP_DIR%\pyarrow\arrow_python_flight.dll exit 1              # [win]
+        - if not exist %SP_DIR%\pyarrow\arrow_python_parquet_encryption.dll exit 1  # [win]
```
h-vetinari, line 899: Reduce to libraries that aren't already in `pyarrow` (and adapt comment)

recipe/meta.yaml
```diff
@@ -747,7 +947,7 @@ outputs:
         # see https://github.com/apache/arrow/issues/37931
         - setuptools_scm <8
       run:
-        - {{ pin_subpackage('pyarrow', exact=True) }}
+        - {{ pin_subpackage('pyarrow-core', exact=True) }}
```
h-vetinari, line 950: This is interesting - i.e. the test suite passes without all underlying libraries available - but I think we should go the opposite way. We should test with `pyarrow-all`, to ensure maximum coverage of the test suite here.

h-vetinari (commented):
Looking pretty good already

recipe/meta.yaml
raulcd: Thanks! yeah, I missed macOS

recipe/meta.yaml
raulcd: yup, I missed the `${SHLIB_EXT}`. Thanks!

recipe/meta.yaml
raulcd: At the moment all `pyarrow-core`, `pyarrow` and `pyarrow-all` contain `libarrow_python`, `libarrow_python_flight` and `libarrow_python_parquet_encryption`. I've removed the test commands for `pyarrow-all` as the files are present on the rest. We probably should remove them for `pyarrow` as they are tested on `pyarrow-core` or we might want to create an `install-pyarrow.sh` script that picks the specific libraries for each one but I am not sure if the effort is worth it.
To be honest the big libraries are the Arrow CPP ones, the ones that we could filter on pyarrow (`arrow_python_flight` and `arrow_python_parquet_encryption`) on 15.0.0 are some KB size:
```
$ ls -lhrt libarrow*so
-rwxrwxr-x 1 raulcd raulcd 115K feb  5 15:50 libarrow_python_flight.so
-rwxrwxr-x 1 raulcd raulcd  37K feb  5 15:50 libarrow_python_parquet_encryption.so
-rwxrwxr-x 1 raulcd raulcd 2,6M feb  5 15:50 libarrow_python.so
```

recipe/meta.yaml
raulcd: Done

h-vetinari (approved):

h-vetinari:
@xhochy, please re-review (or dismiss your open request for changes). :)

jorisvandenbossche:
> I am not entirely sure if we should also remove `libarrow-substrait` [from `pyarrow`]

On the one hand, it seems not necessary to include to keep the majority of use cases working (i.e. not to break too many users that now rely on the `"pyarrow"` conda package). 
On the other hand, if we would decide to keep for example the GCS filesystem in "pyarrow", then size is not really an isue (as GCS also needs libprotobuf, and so that would be included in the env anyway)

recipe/meta.yaml
```diff
@@ -644,8 +658,168 @@ outputs:
         - numpy                                  # [build_platform != target_platform]
         - cmake
         - ninja
+      host:
+        - {{ pin_subpackage("libarrow", exact=True) }}
+        - {{ pin_subpackage("libarrow-acero", exact=True) }}
```
xhochy, line 663: Can you add a comment on why there are listed in `host` but not in `run`? It looks like you did this intentionally, but it would be better if we recorded this decision.

xhochy (changes requested):

recipe/meta.yaml
h-vetinari: @raulcd, would you like to formulate a response to this request? Otherwise I can probably come up with something.

raulcd: I am currently traveling. I will be able to get back to this around Friday, I am happy if you can come up with something before that, otherwise I'll do it then.

recipe/meta.yaml
```diff
@@ -651,8 +665,168 @@ outputs:
         - numpy                                  # [build_platform != target_platform]
         - cmake
         - ninja
+      host:
+        - {{ pin_subpackage("libarrow", exact=True) }}
```
raulcd, line 669: @h-vetinari @xhochy that's my understanding on why they have to be on host, let me know your thoughts:
```suggestion
      host:
        # We add all libarrow package dependencies on host in order to build pyarrow
        # once with all capabilities. We require the exact hash versions but we don't
        # want those to be installed when running for pyarrow-core as we want a
        # low memory footprint for the pyarrow-core package.
        # We cannot use them on build as we want the exact hash version built on the 
        # recipe to be linked, otherwise we get mismatching package errors.
        - {{ pin_subpackage("libarrow", exact=True) }}
```
as an alternative we can add them as build dependencies with:
```diff
diff --git a/recipe/meta.yaml b/recipe/meta.yaml
index 749176b..29ba9fb 100644
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -665,15 +665,15 @@ outputs:
         - numpy                                  # [build_platform != target_platform]
         - cmake
         - ninja
+        - {{ pin_subpackage("libarrow-acero") }}
+        - {{ pin_subpackage("libarrow-dataset") }}
+        - {{ pin_subpackage("libarrow-flight") }}
+        - {{ pin_subpackage("libarrow-flight-sql") }}
+        - {{ pin_subpackage("libarrow-gandiva") }}
+        - {{ pin_subpackage("libarrow-substrait") }}
+        - {{ pin_subpackage("libparquet") }}
       host:
         - {{ pin_subpackage("libarrow", exact=True) }}
-        - {{ pin_subpackage("libarrow-acero", exact=True) }}
-        - {{ pin_subpackage("libarrow-dataset", exact=True) }}
-        - {{ pin_subpackage("libarrow-flight", exact=True) }}
-        - {{ pin_subpackage("libarrow-flight-sql", exact=True) }}
-        - {{ pin_subpackage("libarrow-gandiva", exact=True) }}
-        - {{ pin_subpackage("libarrow-substrait", exact=True) }}
-        - {{ pin_subpackage("libparquet", exact=True) }}
         - clangdev {{ llvm_version }}
         - llvmdev {{ llvm_version }}
         - cython
```
but from my understanding we require `exact=True` to match the one build on the recipe but again, my conda knowledge is limited and I might be doing something wrong here.

h-vetinari:

recipe/meta.yaml
h-vetinari: No, they shouldn't be in build. But the main part you already answered. It's about keeping the footprint of `pyarrow` small - which now also has graceful fallbacks in case the libs aren't found - and leaving the full-fledged install to `pyarrow-all`.

recipe/meta.yaml
raulcd: I've added the comment @xhochy let me know if that makes sense and are you ok with me merging this

raulcd:
I would like to merge this before 15.0.1 is released. @xhochy are you ok with it?

jorisvandenbossche:
I was thinking, given this is somewhat a "breaking" change (for people using flight or gandiva), should we only start doing this with 16.0?

(or already merge, but only remove those dependencies from the default "pyarrow" target starting with 16.0)

h-vetinari:
> I was thinking, given this is somewhat a "breaking" change (for people using flight or gandiva), should we only start doing this with 16.0?

That's a good point. +1 from me

ianmcook:
@jorisvandenbossche @raulcd I think we need to have a discussion about this on the Arrow mailing list before we push this change. The Arrow community currently has little awareness that this change is planned. I worry that there could be much surprise and frustration if we do not inform the community in advance.

raulcd:
@ianmcook I think this is a valid concern but the only affected users will be those using `flight`, `flight-sql` and `gandiva` on pyarrow. When installing they will get a message like:
```
The pyarrow installation is not built with support for 'flight'
```
We could add a patch like:
```diff
diff --git a/python/pyarrow/flight.py b/python/pyarrow/flight.py
index b183690..12eef59 100644
--- a/python/pyarrow/flight.py
+++ b/python/pyarrow/flight.py
@@ -65,5 +65,6 @@ try:
     )
 except ImportError as exc:
     raise ImportError(
-        f"The pyarrow installation is not built with support for 'flight' ({str(exc)})"
+        "The 'pyarrow' installation is not built with support for " +
+        f"'flight'. Please install the conda-forge 'pyarrow-all' package. ({str(exc)})"
     ) from None
```
and users will get prompted something like:
```
The pyarrow installation is not built with support for 'flight'.  Please install the conda-forge 'pyarrow-all' package.
```
to be even more explicit.

We can send an email to the mailing list in order to give a heads up but in my opinion we could merge. We did the split of `libarrow` and `libarrow-all` in the past and didn't cause issues.

ianmcook:
Thanks @raulcd — a message to the user@ and dev@ mailing lists would be great. The patch to give a more helpful error message would also be great.

recipe/meta.yaml
```diff
@@ -764,7 +928,7 @@ outputs:
         - ninja
       host:
         - {{ pin_subpackage("libarrow-all", exact=True) }}
-        - {{ pin_subpackage('pyarrow', exact=True) }}
+        - {{ pin_subpackage('pyarrow-core', exact=True) }}
```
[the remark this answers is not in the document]
raulcd: I think you are right! Thanks!

raulcd:
@h-vetinari I was having a chat with @xhochy at PyCon DE and decided to add the Arrow update to v16.0.0 in the same PR so there is not a build without the split. I hope this is also good with you.

h-vetinari:
That's exactly how I would have suggested doing it! :)

raulcd:
It seems we hit this issue: https://github.com/protocolbuffers/protobuf/issues/14576
trying the workaround proposed there by @h-vetinari

xhochy (approved):
Let's go!

recipe/meta.yaml
```diff
@@ -670,8 +680,174 @@ outputs:
...
+        - {{ pin_subpackage("libparquet", exact=True) }}
+        - {{ pin_subpackage("pyarrow-core", exact=True) }}
+        - clangdev {{ llvm_version }}
+        - llvmdev {{ llvm_version }}
+        - cython
+        - numpy
+        - python
+        - setuptools
+        # see https://github.com/apache/arrow/issues/37931
+        - setuptools_scm <8
```
h-vetinari, line 779: the new `pyarrow` doesn't build anything, it just collects dependencies; so we can completely remove the `host:` section here. Same for `pyarrow-all`.

recipe/meta.yaml
```diff
@@ -670,8 +680,174 @@ outputs:
...
+        - pyarrow.parquet
+        - pyarrow.fs
+        - pyarrow._s3fs
+        - pyarrow._hdfs
+        # We can only test importing cuda package but cannot run when a
+        # CUDA device is not available, for instance, when building from CI.
+        # On Windows, we cannot even do that due to `nvcuda.dll` not being found, see
+        # https://conda-forge.org/docs/maintainer/knowledge_base.html#nvcuda-dll-cannot-be-found-on-windows
+        # However, we check below for (at least) the presence of a correctly-compiled module
+        - pyarrow.cuda     # [cuda_compiler_version != "None" and not win]
```
h-vetinari, line 811: This is redundant with the tests for `pyarrow-core` and needs to be deleted.

recipe/meta.yaml
```diff
@@ -670,8 +680,174 @@ outputs:
...
+        - test ! -f ${SP_DIR}/pyarrow/tests/test_array.py                           # [unix]
+        - if exist %SP_DIR%/pyarrow/tests/test_array.py exit 1                      # [win]
+
+        # Expected not included libraries
+        - test ! -f $PREFIX/lib/libarrow_flight.${SHLIB_EXT}                        # [unix]
+        - test ! -f $PREFIX/lib/libgandiva.${SHLIB_EXT}                             # [unix]
+
+        # Need to remove dot from PY_VER; %MYVAR:x=y% replaces "x" in %MYVAR% with "y"
+        - if not exist %SP_DIR%/pyarrow/_cuda.cp%PY_VER:.=%-win_amd64.pyd exit 1  # [win and cuda_compiler_version != "None"]
+        - python test_read_parquet.py
```
h-vetinari, line 836: Same here. We might want to keep some absolutely minimal testing if only to shut up the linter, but there's nothing to test here, as we've already validated that all the dependent outputs contain what they should

The only real thing to test here is whether the upstream fallbacks for (un)available library artefacts work or not, but arguably, we can just trust upstream arrow on that, because we don't change anything about that mechanism.

recipe/meta.yaml
```diff
@@ -670,8 +680,174 @@ outputs:
...
+
+  - name: pyarrow-all
+    version: {{ version }}
+    requirements:
       host:
         - {{ pin_subpackage("libarrow-all", exact=True) }}
+        - {{ pin_subpackage("pyarrow-core", exact=True) }}
         - clangdev {{ llvm_version }}
         - llvmdev {{ llvm_version }}
         - zlib
```
h-vetinari, line 853: As mentioned above for `pyarrow`, the host section needs to go here as well.

recipe/meta.yaml
```diff
@@ -690,6 +866,7 @@ outputs:
         - {{ pin_subpackage("libarrow-gandiva", exact=True) }}
         - {{ pin_subpackage("libarrow-substrait", exact=True) }}
         - {{ pin_subpackage("libparquet", exact=True) }}
+        - {{ pin_subpackage("pyarrow-core", exact=True) }}
```
h-vetinari, line 869: I'd prefer for this to depend on 
```suggestion
        - {{ pin_subpackage("pyarrow", exact=True) }}
```
That way, it's obvious that there's no "hole" in the layering.

recipe/meta.yaml
```diff
@@ -715,30 +892,13 @@ outputs:
         # https://conda-forge.org/docs/maintainer/knowledge_base.html#nvcuda-dll-cannot-be-found-on-windows
         # However, we check below for (at least) the presence of a correctly-compiled module
         - pyarrow.cuda     # [cuda_compiler_version != "None" and not win]
```
h-vetinari, line 894: Same here, no tests that we've already done, unless they capture a meaningful change between the various different layers.

h-vetinari (changes requested):
This is getting close but still needs a bit of work. I also want to do #1375 first to uncouple the aws-migrations from the arrow bump. I'm happy to rebase this PR on top afterwards (commit history needs some cleanup as well), and apply the necessary clean-ups.

Also, this hasn't been rerendered yet for the new outputs, which is going to further increase the rerender time and effectively break the bot from opening new migrations (see https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/5815). This is not an issue with this PR itself, but something we should fix beforehand, or at the very least ASAP afterwards.

recipe/meta.yaml
raulcd: I've kept some really minor testing but have removed the unnecessary repeated ones.

recipe/meta.yaml
raulcd: Done

ianmcook:
@raulcd is `pyarrow.compute` included in `pyarrow-core`?

h-vetinari:
> @raulcd is `pyarrow.compute` included in `pyarrow-core`?

Mostly yes, see [here](https://github.com/conda-forge/pyarrow-feedstock/blob/64ce0731d0a5e59c0e293c80b5ae7202d30ee021/recipe/meta.yaml#L104-L105):
```yaml
    test:
      imports:
        - pyarrow
        # Compute can be imported but the underlying libarrow_acero is not present.
        - pyarrow.compute
```
I'm not sure how usable it is without acero, that's for @raulcd to tell..

jorisvandenbossche:
The compute kernels itself (`pyarrow.compute`) should be fully usable without acero being present AFAIK

jorisvandenbossche:
(we have an issue about splitting the compute kernels into its own shared library (https://github.com/apache/arrow/issues/25025), and that would allow it here to be installed separately as well, but until then I think the compute kernels are all included (or not, if one would disable building then) in the main libarrow.so)

PR #1293, opened 2024-01-30
[13.x] Rebuild for grpc 1.60 & protobuf 4.25.1

h-vetinari:
The bot cannot open a PR because `cuda120.yaml` is broken as currently checked in.


raulcd:
It seems we might need to backport the pytest pin due to the `pytest-lazy-fixture` issue

PR #1290, opened 2024-01-22
Update to v15.0.0

raulcd:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

PR #1294, opened 2024-01-30
[12.x] Rebuild for grpc 1.60 & protobuf 4.25.1

h-vetinari:
The bot cannot open a PR because `cuda120.yaml` is broken as currently checked in.

### DIFF
--- a/.ci_support/linux_64_cuda_compiler_version11.2.yaml
+++ b/.ci_support/linux_64_cuda_compiler_version11.2.yaml
@@ -35,9 +35,9 @@ google_cloud_cpp:
 libabseil:
 - '20230802'
 libgrpc:
-- '1.59'
+- '1.60'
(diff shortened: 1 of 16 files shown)


raulcd (approved):
All looks good to me. I don't understand when a manual migration is required to be added and is not picked from conda-forge pinning.

h-vetinari:
> I don't understand when a manual migration is required to be added and is not picked from conda-forge pinning.

Due to a change that broke the state of the `cuda120.yaml` migration as currently checked-in on the maintenance branches, the bot ran into the same errors while rerendering that you encountered in #1290. We need 022489c75f6eb6d2de438ecaf92705a57b3555e4 so that things will work again as intended (to be fair, migration number increases are rare; normally fixes to the migration in the global pinning repo automatically get picked up in other feedstocks).

PR #1312, opened 2024-02-19
Preserve symlinks

xhochy:
@h-vetinari I would like to apply this on top of current PRs if you're happy with the change.

Closes #1307


h-vetinari:
Yeah, sounds good to me. We need a fix from #1306 to unblock CI. If you want, I can then rebase #1307 and #1308 into this PR and we fix it in one go.

xhochy:
> Yeah, sounds good to me. We need a fix from #1306 to unblock CI. If you want, I can then rebase #1307 and #1308 into this PR and we fix it in one go.

Yes, that sounds good (and what I thought and thus did not touch the build no).

h-vetinari (approved):
Thank you!

PR #1369, opened 2024-04-18
Backport #40412 to `14.0.2`

galipremsagar:
This PR backports a python memory leak patch to 14.0.2: https://github.com/apache/arrow/pull/40412

Checklist


galipremsagar:

h-vetinari (approved):
Nice!

galipremsagar:
@h-vetinari This might need a CI restart too

galipremsagar:

jakirkham:
Had tried restarting the failing job yesterday evening, but it still fails. Unfortunately the log is sparse on details except for the job stopping at some point. Typically this is a sign of resource limitations 

Guessing this is either too many threads or too little space on the VM. Option to try would be...

1. Reduce [# of threads]( https://github.com/conda-forge/arrow-cpp-feedstock/blob/8639672782d00c711dafa6dab5abd6d2ba917b95/recipe/build.sh#L49-L50 )
2. Freeing more space
3. Using a swap file

More details on 2 & 3 in [this doc]( https://conda-forge.org/docs/maintainer/conda_forge_yml/#azure ), which would involve updating this `conda-forge.yml` section

https://github.com/conda-forge/arrow-cpp-feedstock/blob/8639672782d00c711dafa6dab5abd6d2ba917b95/conda-forge.yml#L1-L2

h-vetinari:
The aarch builds have gotten incredibly flaky recently, but were fine before. Most of the issues happens in the tests, where we're getting errors for fetching something from the web, presumably because the QEMU emulation is too slow. The jobs just straight up dying happens much more rarely, when we get a particularly slow agent. 

> Freeing more space

You're literally pointing out how we're doing that already...? There's no "more" either - it's on or off. I doubt that reducing the threads helps (will just make the build even slower and more prone to timeouts), and I doubt that the memory is the issue, but happy to be proven wrong. 🙃

jakirkham:
Gotcha thanks for the context

Well there are [finer grained settings that do more]( https://github.com/conda-forge/conda-smithy/blob/8a2eba45e40e53abd2e1744475942e133cf6bffe/conda_smithy/templates/azure-pipelines-linux.yml.tmpl#L13-L39 ). They are not turned on by default as they can be a bit slower and in a couple cases builds needed the components removed

That said, no need to spend time doing that if we don't think it will help. Just think up options

h-vetinari:
This has become a reproducible failure at
```
[134/139] Building CXX object CMakeFiles/gandiva.dir/gandiva.cpp.o
[135/139] Linking CXX shared module release/gandiva.cpython-311-aarch64-linux-gnu.so
```
for both v14 & v15 (though not v12 & v13).

h-vetinari:
It's a GCC 12 problem. The CUDA build passes because it uses GCC 10; Downgrading the non-CUDA job to GCC 10 would be possible, but I've also tested that GCC 13 works and prefer that. I'll finish this off with the libgoogle migration, and will come back to this afterwards.

h-vetinari (approved):

galipremsagar:
Thanks @h-vetinari !

PR #1376, opened 2024-04-29
Create subsets of pyarrow package with pyarrow-core < pyarrow < pyarrow-all and update to Arrow v16.0.0

raulcd:
Checklist

Continuation of #1255


recipe/meta.yaml
```diff
@@ -774,8 +873,9 @@ outputs:
         - setuptools
         - setuptools-scm
       run:
-        - {{ pin_subpackage('pyarrow', exact=True) }}
-        - python
+        - {{ pin_subpackage('pyarrow-all', exact=True) }}
+        - {{ pin_compatible('numpy') }}
+        - {{ pin_compatible('python') }}
```
raulcd, line 878: @h-vetinari this is where I get confused on how conda gets the dependencies and solves the environment. I am not sure why but now that the `host` section is removed from `pyarrow-all` I had to add the `pin_compatible` to numpy and python otherwise it uses a different version than the one used to build. Example of the python error:
```
Could not solve for environment specs
The following packages are incompatible
├─ pyarrow-all 16.0.0.* ha770c72_0 is installable and it requires
│  └─ pyarrow 16.0.0 ha770c72_0, which requires
│     └─ pyarrow-core 16.0.0 py39he338278_0_cuda, which requires
│        └─ python_abi 3.9.* *_cp39, which can be installed;
└─ python 3.12.3.* hab00c5b_0_cpython is not installable because it requires
   └─ python_abi 3.12.* *_cp312, which conflicts with any installable versions previously reported.
```
I am not sure if this is the correct solution.

raulcd: Well, this clearly doesn't fix the issue:
```
Could not solve for environment specs
The following packages are incompatible
├─ numpy 1.22.4.* py39hc58783e_0 is requested and can be installed;
└─ pyarrow-all 16.0.0.* ha770c72_0 is not installable because it requires
   └─ pyarrow 16.0.0 ha770c72_0, which requires
      └─ pyarrow-core 16.0.0 py311h225a1a7_0_cuda, which requires
         └─ numpy >=1.23.5,<2.0a0 , which conflicts with any installable versions previously reported.
```

h-vetinari: Ah, OK, yeah, we should still use it, but only where we need the run-exports. E.g.
```
  host:
      # only necessary for run-exports
      - python
      - numpy
```

recipe/meta.yaml
```diff
@@ -671,17 +681,12 @@ outputs:
...
-        - setuptools
-        - setuptools-scm
-      run:
-        # full set of libs because run-exports from libarrow-all aren't picked up
+        # We add all libarrow package dependencies on host in order to build pyarrow
+        # once with all capabilities. We require the exact hash versions but we don't
+        # want those to be installed when running for pyarrow-core as we want a
+        # low memory footprint for the pyarrow-core package.
+        # We cannot use them on build as we want the exact hash version built on the
+        # recipe to be linked, otherwise we get mismatching package errors.
```
h-vetinari, line 689: I don't understand this part of the comment

recipe/meta.yaml
```diff
@@ -690,22 +695,27 @@ outputs:
         - {{ pin_subpackage("libarrow-gandiva", exact=True) }}
         - {{ pin_subpackage("libarrow-substrait", exact=True) }}
         - {{ pin_subpackage("libparquet", exact=True) }}
```
h-vetinari, line 697: Why don't we just keep building against `libarrow-all` here (as before)?

recipe/meta.yaml
```diff
@@ -690,22 +695,27 @@ outputs:
...
+        - clangdev {{ llvm_version }}
+        - llvmdev {{ llvm_version }}
+        - cython
+        - numpy
+        - python
+        - setuptools
+        # see https://github.com/apache/arrow/issues/37931
+        - setuptools_scm <8
```
h-vetinari, line 705: That issue should have been solved for v16, it would be good to remove the cap (also for other outputs)
```suggestion
        - setuptools_scm
```

recipe/meta.yaml
```diff
@@ -717,28 +727,125 @@ outputs:
...
+      imports:
+        # pyarrow-all contains parquet
+        - pyarrow.dataset
+        - pyarrow.parquet
+      commands:
+        # Expected included libraries
+        - test -f $PREFIX/lib/libarrow_acero${SHLIB_EXT}      # [unix]
+        - test -f $PREFIX/lib/libarrow_dataset${SHLIB_EXT}    # [unix]
+        - test -f $PREFIX/lib/libparquet${SHLIB_EXT}          # [unix]
+
```
h-vetinari, line 795: We know they'll be present because we're depending on outputs where we've tested that the respective libraries are there.
```suggestion
```

recipe/meta.yaml
```diff
@@ -717,28 +727,125 @@ outputs:
...
+        # Expected not included libraries
+        - test ! -f $PREFIX/lib/libarrow_flight${SHLIB_EXT}                        # [unix]
+        - test ! -f $PREFIX/lib/libgandiva${SHLIB_EXT}                             # [unix]
```
h-vetinari, line 798: This on the other hand is fine to keep it as a safety for ensuring we don't mess up our pyarrow layering (e.g. by adding the wrong dependencies).

recipe/meta.yaml
```diff
@@ -717,28 +727,125 @@ outputs:
...
+        - {{ pin_subpackage("libparquet", exact=True) }}
+        - {{ pin_subpackage("pyarrow", exact=True) }}
+        - {{ pin_compatible('numpy') }}
+        - python
+      run_constrained:
+        - apache-arrow-proc =*={{ build_ext }}
+
+    test:
+      files:
+        - test_read_parquet.py
```
h-vetinari, line 835: Already tested in `pyarrow`
```suggestion
```

recipe/meta.yaml
```diff
@@ -717,28 +727,125 @@ outputs:
...
+      imports:
+        - pyarrow.flight
+        - pyarrow.gandiva
+      commands:
+        # Expected included libraries
+        - test -f $PREFIX/lib/libarrow_flight${SHLIB_EXT}                        # [unix]
+        - test -f $PREFIX/lib/libgandiva${SHLIB_EXT}                             # [unix]
```
h-vetinari, line 842: Unnecessary
```suggestion
```

h-vetinari (commented):
This is getting close, thanks a lot! Just some minor fixes left.

We should also revert https://github.com/conda-forge/arrow-cpp-feedstock/commit/efd67842e74d241b480d15bf807dea304a1f054a & https://github.com/conda-forge/arrow-cpp-feedstock/commit/985aeec6f23abf6d488de33af72acc338346f5cb (already on 15.x, but I can cherry-pick those reverts if we end up merging this PR first) and remove the comments referring to the caps.

recipe/meta.yaml
raulcd: I was trying to explain why they are on the host section and not on the build section. To be honest I added the comment because I tried to add them on the build section instead of host and I was getting a mismatching package error. I am removing the comment because it is not clear at all and I am not sure how to explain it better. Please do let me know if you think I should add something about that.

recipe/meta.yaml
raulcd: I am not sure what you mean with the `as before` we are already building against all of them because we had some issues with the `run-exports` were we had to define all of the,

h-vetinari: Before this PR, `pyarrow` had a host-dependency on `libarrow-all`:
https://github.com/conda-forge/arrow-cpp-feedstock/blob/a2561f808c8c794139272855ff46ba4ddbdb1e87/recipe/meta.yaml#L673-L674

The comment explicitly says that we still want all the host-dependencies, while not having all the runtime deps (for `pyarrow-core`). So my question was: why don't we just simply keep host-depending on 
```
- {{ pin_subpackage("libarrow-all", exact=True) }}
```
for `pyarrow-core`?

raulcd: I was confused. I think this was a leftover of some debugging I did. Thanks!

h-vetinari:
This PR is now ready from my side. However, unless there's very concrete urgency, I'd like to wait a couple of days until people have had a chance to respond to #1381, which would simply entail moving all the `python-*` outputs here to https://github.com/conda-forge/pyarrow-feedstock. That's not complicated per se, but it would make the transition much easier IMO if this happened at a version boundary.

h-vetinari:
Ah actually, something is still amiss. We're not creating different hashes between the CUDA & non-CUDA python wrappers. That wouldn't be an issue per se (especially if we do #1381), but it _is_ an issue since we're currently doing `pin_subpackage(..., exact=True)`, which means it will be a random race which of the two python wrappers gets uploaded first (per python version & platform):

| linux-64 non-CUDA | linux-64 CUDA | |
|-|-|-|
| pyarrow-core-16.0.0-py38hc396e17_0_cpu.conda        | pyarrow-core-16.0.0-py38h5350213_0_cuda.conda | distinct ✅ |
| pyarrow-core-16.0.0-py39h38d04b8_0_cpu.conda        | pyarrow-core-16.0.0-py39hb79a567_0_cuda.conda | distinct ✅ |
| pyarrow-core-16.0.0-py310hd207890_0_cpu.conda       | pyarrow-core-16.0.0-py310h40d6bed_0_cuda.conda | distinct ✅ |
| pyarrow-core-16.0.0-py311hd5e4297_0_cpu.conda       | pyarrow-core-16.0.0-py311h373265a_0_cuda.conda | distinct ✅ |
| pyarrow-core-16.0.0-py312h3f82784_0_cpu.conda       | pyarrow-core-16.0.0-py312h0354947_0_cuda.conda | distinct ✅ |
| pyarrow-16.0.0-py38hb563948_0.conda                 | pyarrow-16.0.0-py38hb563948_0.conda | matches ❌ |
| pyarrow-16.0.0-py39h8003fee_0.conda                 | pyarrow-16.0.0-py39h8003fee_0.conda | matches ❌ |
| pyarrow-16.0.0-py310h17c5347_0.conda                | pyarrow-16.0.0-py310h17c5347_0.conda | matches ❌ |
| pyarrow-16.0.0-py311h781c19f_0.conda                | pyarrow-16.0.0-py311h781c19f_0.conda | matches ❌ |
| pyarrow-16.0.0-py312h8da182e_0.conda                | pyarrow-16.0.0-py312h8da182e_0.conda | matches ❌ |
| pyarrow-all-16.0.0-py38hb563948_0.conda             | pyarrow-all-16.0.0-py38hb563948_0.conda | matches ❌ |
| pyarrow-all-16.0.0-py39h8003fee_0.conda             | pyarrow-all-16.0.0-py39h8003fee_0.conda | matches ❌ |
| pyarrow-all-16.0.0-py310h17c5347_0.conda            | pyarrow-all-16.0.0-py310h17c5347_0.conda | matches ❌ |
| pyarrow-all-16.0.0-py311h781c19f_0.conda            | pyarrow-all-16.0.0-py311h781c19f_0.conda | matches ❌ |
| pyarrow-all-16.0.0-py312h8da182e_0.conda            | pyarrow-all-16.0.0-py312h8da182e_0.conda | matches ❌ |
| pyarrow-tests-16.0.0-py38hc396e17_0_cpu.conda       | pyarrow-tests-16.0.0-py38h5350213_0_cuda.conda | distinct ✅ |
| pyarrow-tests-16.0.0-py39h38d04b8_0_cpu.conda       | pyarrow-tests-16.0.0-py39hb79a567_0_cuda.conda | distinct ✅ |
| pyarrow-tests-16.0.0-py310hd207890_0_cpu.conda      | pyarrow-tests-16.0.0-py310h40d6bed_0_cuda.conda | distinct ✅ |
| pyarrow-tests-16.0.0-py311hd5e4297_0_cpu.conda      | pyarrow-tests-16.0.0-py311h373265a_0_cuda.conda | distinct ✅ |
| pyarrow-tests-16.0.0-py312h3f82784_0_cpu.conda      | pyarrow-tests-16.0.0-py312h0354947_0_cuda.conda | distinct ✅ |

h-vetinari:
Ugh, the move from CUDA 11.2 -> 11.8 seems to have caused the aarch builds to hang (CC @jakirkham), and the non-CUDA build on ppc is now crashing in the test suite. No idea if this could be related to pytest 8 somehow, the stacktrace is useless.

raulcd:
> No idea if this could be related to pytest 8 somehow, the stacktrace is useless.

I don't think it is as in my last commit (removing the pytest pin) ppc was successful, there was only a flaky test failure on aarch64

h-vetinari:
> I don't think it is as in my last commit (removing the pytest pin) ppc was successful,

Are you sure? Because https://github.com/conda-forge/arrow-cpp-feedstock/pull/1376/commits/99dbba14d612a7a753eeff2d782e65ff2955c852 came after your last commit.

raulcd:
> Are you sure? Because [99dbba1](https://github.com/conda-forge/arrow-cpp-feedstock/commit/99dbba14d612a7a753eeff2d782e65ff2955c852) came after your last commit.

True, I forgot that I didn't push the revert even though I did it locally. I've done so many tests and changes on this PR that I should give it some time :)

h-vetinari:
Turns out it's related to the compiler version. Both GCC 11 (newly pulled in because CUDA 11.8 allows it) breaks on linux-aarch64 and GCC 12 leads to problems on ppc (like it did on aarch).

h-vetinari (approved):
OK, this is ready for merging now. I'd still like to hear some feedback on #1381, but if we cannot come to a decision there quickly, I'm OK to merge this as-is and deal with the split later.

PR #1380, opened 2024-05-01
[15.x] Rebuild for numpy 2.0

h-vetinari:
Test that https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5790 works for recipes which build all python versions in one go. Also adapt to dropping of CUDA 11.2.


h-vetinari:
This fails because various optional test dependencies aren't available for numpy 2.0 yet, but from the POV of the migrator setup, it works.

h-vetinari:
@conda-forge/arrow-cpp, I've tested backporting `numpy 2.0` compatibility to arrow 15 (generally, the more uniformly we can apply a migration, the better), and had no problems after applying the same patches that were also necessary for arrow 16.

In particular, the whole test suite passes, so I'd prefer to merge this, but wanted to ask for feedback if someone disagrees.

I've also quickly tried this for v14 & v13, but there, we already have a cap on the pandas version to avoid test failures, and that cap is incompatible with a pandas build that's compatible with numpy 2.0, so I wouldn't touch those branches anymore.

xhochy:
This means though, that we cannot proceed with the NumPy 2 migration with downstream feedstocks like https://github.com/conda-forge/turbodbc-feedstock/pull/125 ?

h-vetinari:
> This means though, that we cannot proceed with the NumPy 2 migration with downstream feedstocks like [conda-forge/turbodbc-feedstock#125](https://github.com/conda-forge/turbodbc-feedstock/pull/125) ?

Yeah, feedstocks that depend on both arrow and numpy would either have to skip builds for arrow <15, or would have to do something clever with doubling the build matrix and skipping the unsupported combinations. IMO that's just the nature of the game - i.e. supporting 4 arrow versions at any given time means that a new python version (or now: numpy 2) will not be supported across all versions.

I mean, of course anyone can give it a shot to test how invasive a backport to v14 or v13 would be, and once CI is green, we could decide if we're alright with that. But ex ante I'd say that's just the way it is.

raulcd (approved):
I currently don't have the bandwidth to try the port to v13 and v14 so I agree with @h-vetinari on the changes

PR #1382, opened 2024-05-07
[15.x] Migrate ucx116

xhochy:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #1384, opened 2024-05-08
Pin `compiler`s in `meta.yaml` (15.x)

jakirkham:
Move the `compiler` pinnings to `meta.yaml`. This has a few benefits:

* Smoother re-rendering experiences as `compiler` `zip_keys` change
* No need to reproduce unchanged `compiler` version constraints
* Handle LLVM pinning in `meta.yaml` where the version is set
* Easy to change in one place

That said, this can be a bit more cryptic and verbose than `conda_build_config.yaml`. So there are certainly tradeoffs. Am hoping some of these pins can roll off over time, leading to simpler YAML in the long run.

<hr>

Checklist


recipe/meta.yaml
```diff
@@ -31,18 +31,30 @@ source:
...
   # for cuda support, building with one version is enough to be compatible with
   # all later versions, since arrow is only using libcuda, and not libcudart.
   skip: true  # [cuda_compiler_version not in ("None", cuda_compiler_version_min)]
   # arrow promises API- & ABI-compatibility along SemVer, see #1096
 
 requirements:
   build:
-    - {{ compiler("c") }}
     - {{ stdlib("c") }}
+    {% if cuda_compiler_version == "None" %}
```
jakirkham, line 43: Have grouped the non-CUDA cases together as it seems those are the ones that are mainly being pinned

h-vetinari:
Thanks John. I already adapted to the CUDA 11.2 changes in https://github.com/conda-forge/arrow-cpp-feedstock/commit/201625cb8108179033b65368f6f7b3e2af68de8c, and #1382 is ready to merge as soon as we have ucx-PRs for all branches.

That aside, setting the compiler pins in CBC is well-established practice, and I do not think exploding the compiler-configuration in `meta.yaml` is a worthwhile trade-off. So -1 from me on this.

jakirkham:
Ok wanted to present this option for consideration

Recognize that `zip_keys` can sometimes be hard to work with. We struggled with them for a long time in CUDA packages

Think `zip_keys` presents challenges as we likely lack the proper tooling to modify `zip_keys` in an effective manner (like selecting an entry and replacing certain values or appending or ...)

Ofc this is simply a different kind of workaround to those issues with its own tradeoffs

If we would rather skip this approach, that is ok with me

h-vetinari:
Part of the reason that I don't want to go this way is that compiler bugs in GCC that force us to modify the global pinning are relatively rare. And the clang pins here are strictly speaking not necessary, we could also use `cxx_compiler_version` to set the versions for `llvmdev` & `clangdev` (it's only coupled on osx-arm anyway).

I mainly ended up this way because I didn't want to push the llvm version on maintenance branches when we upgrade compilers - the llvmdev version basically doesn't matter[^1]

[^1]: except when it does - we get segfault with `sparse` through numba due to the version mismatch and not building llvmlite against static libllvm, see https://github.com/conda-forge/llvmlite-feedstock/issues/72

h-vetinari:
Closing this as obsolete. Thanks for the discussion

PR #1385, opened 2024-05-08
[14.x] Migrate ucx116

h-vetinari:


recipe/conda_build_config.yaml
raulcd, line 1: Should we update the comment on the file to explain why we are pinning the compiler version to 10 on CUDA?

Just for my knowledge, how does the order works here, we have:
```
  - 13                         # [linux and aarch64]
  - 10                         # [linux and aarch64]
```
What determines what is used on the CUDA vs non-CUDA?

h-vetinari: We're partially overriding the following [zip](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml#L157-L165) in the global pinning. Since we're only overriding a component, we need to match the [ordering](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/0b245ee178680c357ab81f14fba8fced169d82c1/recipe/conda_build_config.yaml#L6-L9) there, which happens to be non-CUDA and then CUDA (our recipe triggers `CF_CUDA_ENABLED` to be true because we use `{{ compiler("cuda") }}`).

> Should we update the comment on the file to explain why we are pinning the compiler version to 10 on CUDA?

The reason is that arrow v15/v16 hung in compilation when using GCC 11 on aarch+CUDA. We cannot upgrade (because CUDA 11.8 has a maximum supported GCC version of 11), so we downgrade to the last working version.

raulcd: Thanks!

PR #1405, opened 2024-05-15
[15.x] re-enable sparse tests

h-vetinari:
Same as https://github.com/conda-forge/pyarrow-feedstock/pull/121 but for 15.x


h-vetinari:
Included this in #1411

PR #1407, opened 2024-05-15
libarrow 16.1.0

h-vetinari:
Bot failed to open PR for some reason


h-vetinari:
@raulcd, any particular reason why this release bumps the minor version and not the patch version? I believe it's the first such time since v1.0 that the minor version has been bumped?

raulcd:
> any particular reason why this release bumps the minor version and not the patch version?

In general we only do bug fix releases (patch releases) but this one contains not only bug fixes but some features that were requested mainly from the JavaScript implementation. No other reason apart from some minor features included instead of just bug fixes.

h-vetinari:
OK, good to know. I mainly wanted to know if it's something to do with breaking ABI, since we're now only pinning to the major version, but it sounds like it's not an issue.

PR #1443, opened 2024-06-13
[12.x] Rerender to apply latest migrations

jdblischak:
Checklist

---

This PR attempts to rerender the 12.x branch to migrate arrow 12 to the latest conda-forge pinnings

I know that the 12.x branch is no longer supported, so I appreciate that this may fail

https://github.com/conda-forge/arrow-cpp-feedstock/blob/2f976c55060605a8f6ff265d07e5ce55dad8231b/conda-forge.yml#L4-L8

Motivation: We're stuck on pyarrow/libarrow 12 because of some obscure S3 error that only occurs on macOS (https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/171#issuecomment-2166131299)


recipe/conda_build_config.yaml
```diff
@@ -4,3 +4,19 @@ c_compiler_version:            # [osx]
   - 15                         # [osx]
 cxx_compiler_version:          # [osx]
   - 15                         # [osx]
+# Do not try to build for Python 3.12
```
h-vetinari, line 7: Don't override the CBC, simply add `skip: true  # [py==312]` to the python-related outputs.

jdblischak: I can't get the `skip` strategy to work

* That was my first instinct yesterday. I added `skip: true  # [py < 312]` to the top-level `build` section. But rerendering still added Python 3.12
* From your feedback, today I tried putting `skip: true  # [py==312]` in the `build` sections for the outputs `pyarrow` and `pyarrow-tests`. Rerendering again added back Python 3.12
* I also tried combining these, putting `skip: true  # [py==312]` in the top-level `build` section and the Python outputs. Python 3.12 was still added
* Lastly, I tried adding `skip: true  # [py==312]` to the top-level `build` section as well as the `build` section of every single one of the outputs. Python 3.12 was still added

I'm open to using an alternative strategy instead of directly editing the CBC, but I need some assistance figuring out how to get this multiple output recipe to recognize the `skip`

h-vetinari: > I'm open to using an alternative strategy instead of directly editing the CBC, but I need some assistance figuring out how to get this multiple output recipe to recognize the `skip`

There are two different things at play:
* what smithy puts in the variant configs
* what actually gets build based on the skips per output

You're concerning yourself with what values are present after rerendering in the `.ci_support/*yaml` files, but it's irrelevant; as long as the skip for `pyarrow{,-tests}` is correct, nothing will be built for 3.12.

h-vetinari (approved):

recipe/conda_build_config.yaml
jdblischak: > You're concerning yourself with what values are present after rerendering in the `.ci_support/*yaml` files, but it's irrelevant; as long as the skip for `pyarrow{,-tests}` is correct, nothing will be built for 3.12.

@h-vetinari Ah, thanks so much for clearing up my confusion. Now I understand. And thanks for cleaning up my PR and merging!

PR #1431, opened 2024-05-24
Add Dependencies for Arrow's Azure Filesystem

srilman:
Resolved #1430

Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #1452, opened 2024-06-24
Support azure bindings also on windows

h-vetinari:
Retry windows-support that got dropped in #1431


h-vetinari:
The relevant changes from this (and https://github.com/apache/arrow/pull/48971) were picked into:
* #1919 
* #1920 
* #1921 

The backport to v20 did not immediately compile, so v21 is as far back as I'm willing to make the effort to go. 🙃

PR #1537, opened 2024-09-14
Unpin compilers

h-vetinari:
#1530 unpinned some compilers successfully, lets see if this also works on main and for the rest.


h-vetinari:
I've done this in to #1565 now.

PR #1641, opened 2024-12-08
[15.x] migration catch-up

h-vetinari:
A few migrations fell off the wagon for 15.x recently


h-vetinari:
Also running into https://github.com/conda-forge/lz4-c-feedstock/issues/36

PR #1432, opened 2024-05-27
TEST: arrow v18.0.0rc0

h-vetinari:
Some changes require testing across both libarrow and pyarrow; since we split off the pyarrow builds in a66c7c666f1ca3ee1f4ffcf825e9f61d2d6bd77b, we'd need to publish a changed `libarrow`, test the changes in pyarrow, and then potentially mark the builds as broken. In order to simplify this process, reintroduce some end-to-end testing here that can be used to do this all in one PR.

PR #1664, opened 2025-01-11
TEST: arrow v25.0.0 (end-to-end)

h-vetinari:
Some changes require testing across both libarrow and pyarrow; since we split off the pyarrow builds in a66c7c666f1ca3ee1f4ffcf825e9f61d2d6bd77b, we'd need to publish a changed `libarrow`, test the changes in pyarrow, and then potentially mark the builds as broken. In order to simplify this process, reintroduce some end-to-end testing here that can be used to do this all in one PR.

Continued from #1432

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -799,6 +799,318 @@ outputs:
         - LICENSE.txt
       summary: Executables for manipulating Apache arrow files
 
+  - name: pyarrow-core
+    script: build-pyarrow.sh   # [unix]
+    script: build-pyarrow.bat  # [win]
+    version: {{ version }}
+    build:
+      string: py{{ CONDA_PY }}h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}_{{ build_ext }}
+      ignore_run_exports_from:
+        - {{ compiler("cuda") }}                 # [cuda_compiler_version != "None"]
+        - libarrow-all
+        # we don't need numpy at runtime, just to build
+        - numpy
+      rpaths:
+        - lib/
+        - {{ SP_DIR }}/pyarrow
+      missing_dso_whitelist:
+        # not actually missing, but installed into SP_DIR, see tests
+        - '*/arrow_python.dll'                # [win]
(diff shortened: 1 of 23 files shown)


raulcd:
This seems deprecated (as has already been done). I am closing it, feel free to reopen if necessary.

raulcd:
@h-vetinari we have split `libarrow-compute` to its own shared library instead of being bundled with libarrow. We will have to do some work to adapt that and the dependencies:
- https://github.com/apache/arrow/pull/46261

I was planning to do some work there but I wanted to give you a heads up in case you start working on it

h-vetinari:
Thanks for the heads-up @raulcd, feel free to either push into this PR or open a new one that adapts to the new schema.

I just saw the notifications of the rc tags and decided to take them for a spin. :)

recipe/meta.yaml
```diff
@@ -351,6 +350,57 @@ outputs:
...
+      build:
+        - cmake <4
+        - ninja
+        # for strong run-exports
+        - {{ stdlib("c") }}
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        # needs to run protoc & grpc_cpp_plugin
+        - libgrpc                                # [build_platform != target_platform]
+        - libprotobuf                            # [build_platform != target_platform]
```
raulcd, line 371: are those required? they shouldn't :thinking:

raulcd (commented):
Thanks for working on this! I was at EuroPython last week but I am back now.

libarrow-compute is required for Acero and Dataset, we should add them to their host/run dependencies too.

I am +1 on adding it to pyarrow-core, we haven't done the work for it to be an optional dependency on pyarrow yet.

recipe/meta.yaml
h-vetinari: This was just C&P from some other output. 🙃

h-vetinari:
arrow v25.0.0-rc1 is looking good from the POV of conda-forge! 🥳

CC @raulcd @kou @pitrou @jorisvandenbossche @assignUser

PR #1701, opened 2025-02-11
TEST: trigger a no change build

raulcd:
Just a test for an empty build to validate AWSSDK fails as in other PRs


h-vetinari:
This served its purpose

PR #1748, opened 2025-05-05
Add utility executables

AntoinePrv:
Checklist

Splits and split names suggestions are welcome.


recipe/bld.bat
```diff
@@ -39,6 +39,7 @@ cmake -G "Ninja" ^
       -DARROW_ORC:BOOL=ON ^
       -DARROW_PACKAGE_PREFIX="%LIBRARY_PREFIX%" ^
       -DARROW_PARQUET:BOOL=ON ^
+      -DPARQUET_BUILD_EXECUTABLES:BOOL=ON ^
```
pitrou, line 42: Could also pass `-DARROW_BUILD_UTILITIES=ON` (same in `build.sh`)

AntoinePrv: Good point! I'll wait for the maintainers if they want one or two packages for the parquet/arrow executables.

recipe/meta.yaml
```diff
@@ -651,6 +651,46 @@ outputs:
...
+        - test -f $PREFIX/bin/parquet-dump-arrow-statistics                    # [unix]
+        - test -f $PREFIX/bin/parquet-dump-footer                              # [unix]
+        - test -f $PREFIX/bin/parquet-dump-schema                              # [unix]
+        - test -f $PREFIX/bin/parquet-reader                                   # [unix]
+        - test -f $PREFIX/bin/parquet-scan                                     # [unix]
+        - if not exist %LIBRARY_BIN%\parquet-dump-arrow-statistics.exe exit 1  # [win]
+        - if not exist %LIBRARY_BIN%\parquet-dump-footer.exe exit 1            # [win]
+        - if not exist %LIBRARY_BIN%\parquet-dump-schema.exe exit 1            # [win]
+        - if not exist %LIBRARY_BIN%\parquet-reader.exe exit 1                 # [win]
+        - if not exist %LIBRARY_BIN%\parquet-scan.exe exit 1                   # [win]
```
pitrou, line 685: Ideally we would test that they run successfully on a sample Parquet file (that would help weed out any linkage issue), but I don't know the conda-forge common practice about this. @h-vetinari

h-vetinari: I'm personally always in favour of testing functionality! It would be nice to download the parquet file on the fly though, otherwise it ends up getting shipped with every install (conda packages the stuff that goes into the `test:` section as well).

h-vetinari (commented):
This looks very good to me, thanks! Since when is this supported? Should we backport this further back (we currently support v17-v20)?

W.r.t. naming, we don't have a unified nomenclature for "the binaries of a package"; we have `libfoo`, `libfoo-devel`, (opt. `libfoo-static`) but nothing for `foo-<bin>`. IMO `parquet-utils` is fine. CC @xhochy and the rest of @conda-forge/arrow-cpp if you have opinions.

h-vetinari:
Aside from failing CI of course. ;-)

AntoinePrv:
@h-vetinari we need to add feedstock output:
https://github.com/conda-forge/admin-requests/pull/1513

recipe/meta.yaml
```diff
@@ -651,6 +651,84 @@ outputs:
...
+        - if not exist %LIBRARY_BIN%\arrow-file-to-stream.exe exit 1                      # [win]
+        - if not exist %LIBRARY_BIN%\arrow-stream-to-file.exe exit 1                      # [win]
+        - arrow-file-to-stream 2>&1 | findstr /I "Usage: arrow-file-to-stream" || exit 1  # [win]
+
+    about:
+      home: http://github.com/apache/arrow
+      license: Apache-2.0
+      license_file:
+        - LICENSE.txt
+      summary: C++ executables for inspecting manipulating Apache arrow files
```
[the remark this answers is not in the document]
pitrou: And is it necessary to mention "C++"? They are executables regardless.

recipe/meta.yaml
```diff
@@ -651,6 +651,84 @@ outputs:
...
+        - if not exist %LIBRARY_BIN%\parquet-reader.exe exit 1                 # [win]
+        - if not exist %LIBRARY_BIN%\parquet-scan.exe exit 1                   # [win]
+        - parquet-reader 2>&1 | findstr /I "Usage: parquet-reader" || exit 1   # [win]
+
+    about:
+      home: http://github.com/apache/arrow
+      license: Apache-2.0
+      license_file:
+        - LICENSE.txt
+      summary: C++ executables for inspecting Apache Parquet files
```
pitrou, line 694: Is it necessary to mention "C++" here too?

pitrou:
> This looks very good to me, thanks! Since when is this supported?

Since almost forever, I think.

> Should we backport this further back (we currently support v17-v20)?

It mostly depends on conda-forge's backport policy.

h-vetinari:
> It mostly depends on conda-forge's backport policy.

The closest thing to a policy we have is "we support only the last 4 major versions"; I tend to not do feature backports, but more out of stability reasons, rather than anything else. Adding the binaries is unlikely to break something IMO.

recipe/meta.yaml
```diff
@@ -651,6 +652,80 @@ outputs:
         - LICENSE.txt
       summary: C++ libraries for Apache Parquet
 
+  - name: parquet-utils
```
h-vetinari, line 655: Shouldn't the utils be part of `libarrow-all`? The `-all` is pretty unambiguously maximal IMO

h-vetinari: The `lib` argument is valid, perhaps more than the "`-all` means everything a feedstock provides in one go". 

From my POV, the individual libraries already serve the "I want to compile against _this_ library", so `libarrow-all` is primarily (IMO) just a convenience wrapper.

AntoinePrv: I defer to you

recipe/install-libarrow.sh
```diff
@@ -81,9 +81,6 @@ elif [[ "${PKG_NAME}" == "libarrow-all" ]]; then
     rm -f $PREFIX/lib/{libarrow_testing,libarrow_flight_testing}.*
     rm -f $PREFIX/lib/pkgconfig/{arrow-testing,arrow-flight-testing}.pc
     rm -rf $PREFIX/lib/cmake/{ArrowTesting,ArrowFlightTesting}
-    # Remove executables
-    rm -f $PREFIX/bin/parquet-*
-    rm -f $PREFIX/bin/arrow-*
```
h-vetinari: As I said, if you remove this you need to host-depend on the utils in `libarrow-all` as well.

recipe/meta.yaml
```diff
@@ -165,6 +165,8 @@ outputs:
         - {{ pin_subpackage("libarrow-gandiva", exact=True) }}
         - {{ pin_subpackage("libarrow-substrait", exact=True) }}
         - {{ pin_subpackage("libparquet", exact=True) }}
+        - {{ pin_subpackage("arrow-utils", exact=True) }}
+        - {{ pin_subpackage("parquet-utils", exact=True) }}
```
h-vetinari, line 169: i.e. copy these two lines also to `host:`

h-vetinari (changes requested):

h-vetinari (approved):

h-vetinari:
Thanks for the PR!

h-vetinari:
So I tried backporting this in #1758, #1759 and #1760, but it seems that `parquet-dump-arrow-statistics` was only added in v19. If you can provide the right set of binaries to test for for v18 & v17, I'm happy to merge it. For now I'll just do v19

AntoinePrv:
Thanks! I've left a review you can merge.

PR #1773, opened 2025-06-13
[main] Stay on CUDA 11.8

h-vetinari:
With the removal of CUDA 11.8 as a default version, our `cuda_compiler_version_min` will soon be 12.6 or even 12.9

Given that we have a custom migrator for the [express purpose](https://conda-forge.org/news/2025/05/29/cuda-118/) of staying on CUDA 11.8 where that's necessary or beneficial, and given that @kkraus14 has always advocated for keeping the lowest-possible bound in `__cuda >=X`, let's use this here.

~This currently uses fix-ups from https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/7482 & https://github.com/conda-forge/conda-smithy/pull/2338 to work. Once those things have settled, I'm happy to backport this to the maintenance branches.~

### DIFF
--- a/.ci_support/linux_64_c_compiler_version11cuda_compiler_version11.8cxx_compiler_version11.yaml
+++ b/.ci_support/linux_64_c_compiler_version11cuda_compiler_version11.8cxx_compiler_version11.yaml
@@ -77,8 +77,8 @@ thrift_cpp:
 zip_keys:
 - - c_compiler_version
   - cxx_compiler_version
-  - cuda_compiler
   - cuda_compiler_version
+  - cuda_compiler
   - docker_image
 zlib:
 - '1'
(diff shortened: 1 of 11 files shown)


kkraus14:
@h-vetinari some food for future thought.

If/when we want to move to CUDA 12, we should be able to leverage the `cuda-cudart-dev` (includes the `cuda.h` header for the driver for some reason) and the `cuda-driver-dev` (includes a stub library for the CUDA driver library) packages to build without any need for the cuda compiler or any of the infrastructure surrounding it.

That being said, I really don't know who uses the arrow cuda extensions anymore, as cuDF no longer uses it.

h-vetinari:
> If/when we want to move to CUDA 12, we should be able to leverage the `cuda-cudart-dev` (includes the `cuda.h` header for the driver for some reason) and the `cuda-driver-dev` (includes a stub library for the CUDA driver library) packages to build without any need for the cuda compiler or any of the infrastructure surrounding it.

Sounds very nice! Though IIUC, we'd still have to do separate builds for CUDA/non-CUDA? The best of all worlds would be to have a single build that includes the CUDA-capable bits, but only have those be used (using some kind of dynamic dispatch?) if a GPU is present.

> That being said, I really don't know who uses the arrow cuda extensions anymore, as cuDF no longer uses it.

That's mostly a question for the arrow people, I just help package it 😅 

CC @pitrou @kou @raulcd @jorisvandenbossche @assignUser; also JFYI @xhochy

